PR #74, opened 2023-12-13
Add server API to enumerate Wi-Fi Access Points

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

* Allow clients to discover the access points that are available on the server, including their current status (enabled or not) and capabilities.

### Technical Details

* Add new API `WifiEnumerateAccessPoints` to `NetRemote` protobuf gprc structure which enumerates the access points available on the server, including their enablement status and capabilities.
* Implement `WifiEnumerateAccessPoints` with a stub.
* Add basic unit test for `WifiEnumerateAccessPoints` API.
* Use plog verbose for logging API invocations instead of `std::cout`.

### Test Results

* New unit tests pass.

### Reviewer Focus

* Consider whether this API provides enough actionable information for tests to use.
* Consider whether this API will be extensible for future uses.

### Future Work

* Implement `WifiEnumerateAccessPoints` properly.

### Checklist

### DIFF
--- a/protocol/protos/NetRemoteWifi.proto
+++ b/protocol/protos/NetRemoteWifi.proto
@@ -36,11 +36,11 @@ message WifiConfigureAccessPointResult
     bool Succeeded = 2;
 }
 
-message AccessPointDescriptor
+message WifiEnumerateAccessPointsResultItem
 {
     string AccessPointId = 1;
-    bool IsEnabled = 2;
-    Microsoft.Net.Wifi.AccessPointCapabilities Capabilities = 3;
+    Microsoft.Net.Wifi.AccessPointCapabilities Capabilities = 2;
+    bool IsEnabled = 3;
 }
 
 message WifiEnumerateAccessPointsRequest
@@ -49,5 +49,5 @@ message WifiEnumerateAccessPointsRequest
 
 message WifiEnumerateAccessPointsResult
 {
-    repeated Microsoft.Net.Remote.Wifi.AccessPointDescriptor AccessPoints = 1;
+    repeated Microsoft.Net.Remote.Wifi.WifiEnumerateAccessPointsResultItem AccessPoints = 1;
 }
--- a/protocol/protos/NetRemoteWifi.proto
+++ b/protocol/protos/NetRemoteWifi.proto
@@ -35,3 +35,12 @@ message WifiConfigureAccessPointResult
     //
     bool Succeeded = 2;
 }
+
+message WifiEnumerateAccessPointsRequest
+{
+}
+
+message WifiEnumerateAccessPointsResult
+{
+
+}
--- a/protocol/protos/NetRemoteWifi.proto
+++ b/protocol/protos/NetRemoteWifi.proto
@@ -36,11 +36,18 @@ message WifiConfigureAccessPointResult
     bool Succeeded = 2;
 }
 
+message AccessPointDescriptor
+{
+    string AccessPointId = 1;
+    bool IsEnabled = 2;
+    Microsoft.Net.Wifi.AccessPointCapabilities Capabilities = 3;
+}
+
 message WifiEnumerateAccessPointsRequest
 {
 }
 
 message WifiEnumerateAccessPointsResult
 {
-
+    repeated Microsoft.Net.Remote.Wifi.AccessPointDescriptor AccessPoints = 1;
 }
--- a/protocol/protos/WifiCore.proto
+++ b/protocol/protos/WifiCore.proto
@@ -92,7 +92,8 @@ message AccessPointConfiguration
 
 message AccessPointCapabilities
 {
-    repeated Microsoft.Net.Wifi.RadioBand SupportedBands = 1;
+    repeated Microsoft.Net.Wifi.RadioBand Bands = 1;
+    repeated Microsoft.Net.Wifi.Dot11AuthenticationAlgorithm AuthenticationAlgorithms = 2;
 }
 
 enum AccessPointState
--- a/protocol/protos/WifiCore.proto
+++ b/protocol/protos/WifiCore.proto
@@ -89,3 +89,13 @@ message AccessPointConfiguration
 {
     Dot11PhyType PhyType = 1;
 }
+
+message AccessPointCapabilities
+{
+    repeated Microsoft.Net.Wifi.RadioBand SupportedBands = 1;
+}
+
+enum AccessPointState
+{
+    AccessPointStateUnknown = 0;
+}
--- a/protocol/protos/WifiCore.proto
+++ b/protocol/protos/WifiCore.proto
@@ -93,7 +93,8 @@ message AccessPointConfiguration
 message AccessPointCapabilities
 {
     repeated Microsoft.Net.Wifi.RadioBand Bands = 1;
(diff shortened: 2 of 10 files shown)


protocol/protos/WifiCore.proto
```diff
@@ -89,3 +89,13 @@ message AccessPointConfiguration
...
 }
+
+message AccessPointCapabilities
+{
+    repeated Microsoft.Net.Wifi.RadioBand SupportedBands = 1;
+}
+
+enum AccessPointState
+{
+    AccessPointStateUnknown = 0;
```
corbin-phipps, line 100: Do we need other states here?

src/common/server/NetRemoteServer.cxx
```diff
@@ -34,7 +35,7 @@ void NetRemoteServer::Run()
     builder.RegisterService(&m_service);
 
     m_server = builder.BuildAndStart();
-    std::cout << std::format("Started listening on {}", m_serverAddress) << std::endl;
+    LOG_INFO << std::format("netremote server started listening on {}", m_serverAddress) << std::endl;
```
corbin-phipps, line 38: I don't think we need `std::endl` here anymore for plog

protocol/protos/WifiCore.proto
```diff
@@ -89,3 +89,13 @@ message AccessPointConfiguration
 {
     Dot11PhyType PhyType = 1;
 }
+
+message AccessPointCapabilities
+{
+    repeated Microsoft.Net.Wifi.RadioBand SupportedBands = 1;
+}
+
+enum AccessPointState
```
corbin-phipps, line 98: Is this used anywhere? Maybe should be part of `AccessPointDescriptor`?

protocol/protos/WifiCore.proto
abeltrano: Yes, definitely, I just don't know exactly what they'll be yet.

protocol/protos/WifiCore.proto
abeltrano: It's not used anywhere yet. I initially thought it would go in `AccessPointDescriptor`, but that is intended to describe the static, unchanging properties of an access point. So, if you think that makes sense, then I think the question would be should `AccessPointState` be included in the `WifiEnumerateAccessPointsResult` for each access point? For example, instead of providing a list of `AccessPointDescriptor`s, we'd instead do something like:

```proto
message WifiEnumerateAccessPointsRequestResultEntry
{
    AccessPointDescriptor Descriptor;
    AccessPointState State;
}

message WifiEnumerateAccessPointsResult
{
    repeated WifiEnumerateAccessPointsRequestResultEntry AccessPoints = 1;
}
```

We would need to query each enabled AP to get the state. That makes this AP more active, and the idea was to make it something the clients could use simply to see what APs were available and their capabilities. So, my initial thought is not to include it and instead, have a separate API that allows interaction with the AP to do active queries/actions with it, like querying its current state, changing certain settings, etc.

Thoughts?

protocol/protos/WifiCore.proto
```diff
@@ -89,3 +89,13 @@ message AccessPointConfiguration
 {
     Dot11PhyType PhyType = 1;
 }
+
+message AccessPointCapabilities
```
corbin-phipps, line 93: Should this also contain the various dot11 algos as fields?

protocol/protos/WifiCore.proto
corbin-phipps: Oh I see, I think it makes sense to leave `AccessPointDescriptor` as something that contains the unchanging properties, although could `IsEnabled` be changed?

I think your initial idea to have a separate API to query info about the AP like its current state is a good idea, so no need to change anything here

protocol/protos/WifiCore.proto
abeltrano: I think so, but I'm not totally certain what mix of things will need to go here. For example, do both the algorithms _and_ the ciphers need to go here? What about the phy type? 

I initially decided to defer this decision for another PR so I could think it through more, but now that you've raised it, I don't see any case where we wouldn't provide the algorithms, so, I'll add a field here for that.

src/common/server/NetRemoteServer.cxx
abeltrano: Removed.

protocol/protos/WifiCore.proto
abeltrano: Added.

protocol/protos/WifiCore.proto
abeltrano: Great point about `IsEnabled`, it shouldn't be in there. I'll factor it out.

abeltrano: I ended up removing `AccessPointDescriptor` all together because it didn't make sense to exist on its own. To that end, I moved all of the result arguments into their own message `WifiEnumerateAccessPointResultItem`.

protocol/protos/NetRemoteWifi.proto
```diff
@@ -35,3 +35,19 @@ message WifiConfigureAccessPointResult
...
+    bool IsEnabled = 3;
+}
+
+message WifiEnumerateAccessPointsRequest
+{
+}
+
+message WifiEnumerateAccessPointsResult
+{
+    repeated Microsoft.Net.Remote.Wifi.WifiEnumerateAccessPointsResultItem AccessPoints = 1;
```
abeltrano, line 52: FYI @corbin-phipps, I opted to use a list of `WifiEnumerateAccessPointsResultItem`s which includes the `AccessPointId` instead of `map<string, WifiEnumerateAccessPointsResultItem>` because the client would presumably never actually do a lookup on the map (which is what maps/dictionaries are designed for), but instead would immediately just iterate or filter it for the properties it wants, which is inherently more list-like.

This then raises the question of whether we should add something to `WifiEnumerateAccessPointsRequest` which allows filtering the results with the condition(s) it's looking for. Given we expect this list to be relatively small (on the order of 10s at absolute _worst_), it didn't seem appropriate to complicate the API and implementation with filtering options.

What do you think?

tests/unit/TestNetRemoteServiceClient.cxx
```diff
@@ -47,3 +48,43 @@ TEST_CASE("WifiConfigureAccessPoint API can be called", "[basic][rpc][client][re
...
+
+        WifiEnumerateAccessPointsResult result{};
+        grpc::ClientContext clientContext{};
+
+        auto status = client->WifiEnumerateAccessPoints(&clientContext, request, &result);
+        REQUIRE(status.ok());
+        // for (const auto& accessPoint : result.accesspoints())
+        // {
+        //     REQUIRE(accessPoint.enabled() == false);
+        // }
```
abeltrano, line 88: Oops, will update.

protocol/protos/NetRemoteWifi.proto
corbin-phipps: I think leaving the filtering to the client is fine. In general, I think simpler APIs with fewer options tend to be best and makes them easier to maintain

corbin-phipps (approved):

PR #76, opened 2023-12-15
Proposed API for field-updatable ap configuration.

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [X] Breaking change
- [ ] Non-functional change

### Goals

* Provide an API for updating many configuration settings of an access point at once. 
    - The driving scenario is to allow an initial, base/core set of configurations to be applied to an access point when initially enabled.

### Technical Details

*  This proposes an API which enables partial access point configuration updates in a single call. The request structure `WifiAccessPointApplyConfigurationRequest` takes in series of configurations, one for each active radio band, and attempts to apply these configurations. 

### Test Results

* All added unit tests pass.

### Reviewer Focus

* Consider whether partial configuration in a single API is needed at all, compared to the alternative of providing distinct APIs to change configuration options instead, of which some will be required anyway (eg. changing authentication mechanism and the active band are just a few).
* Consider whether this functionality could be achieved differently by requiring a core set of configuration when enabling an access point, something like (roughly):

```proto

rpc WifiAccessPointEnable (WifiAccessPointEnableRequest, WifiAccessPointEnableResult);
message WifiAccessPointEnableRequest
{
    string AccessPointId;
    WifiAccessPointConfigurationCore CoreConfiguration;
}

message Ssid {/* ... */};
message WifiAccessPointConfigurationCore
{
    RadioBand Band;
    Ssid Ssid;
    Dot11PhyType PhyType;
    Dot11AuthenticationAlgorithm AuthenticationAlgorithm;
    // Other core configuration required for basic operation here.
}
```

### Future Work

* Propose alternative API for achieving same effect.

### Checklist

- [ ] Build target `all` compiles cleanly.
- [ ] clang-format and clang-tidy deltas produced no new output.
- [ ] Newly added functions include doxygen-style comment block.

### DIFF
--- a/protocol/protos/NetRemoteWifi.proto
+++ b/protocol/protos/NetRemoteWifi.proto
@@ -51,7 +51,7 @@ message WifiAccessPointApplyConfigurationResultConfigurationItem
     // The status of applying the configuration.
     WifiAccessPointApplyConfigurationStatus Status = 3;
 
-    // Indicates the parts of the corresponging 'WifiAccessPointApplyConfigurationRequestConfigurationItem.Configuration'
+    // Indicates the parts of the corresponding 'WifiAccessPointApplyConfigurationRequestConfigurationItem.Configuration'
     // field that were successfully applied. Set only when Status != WifiAccessPointApplyConfigurationStatusSucceeded.
     google.protobuf.FieldMask ConfigurationFailuresMask = 4;
 }
--- a/protocol/protos/NetRemoteWifi.proto
+++ b/protocol/protos/NetRemoteWifi.proto
@@ -3,37 +3,79 @@ syntax = "proto3";
 
 package Microsoft.Net.Remote.Wifi;
 
+import "google/protobuf/field_mask.proto";
 import "WifiCore.proto";
 
-message WifiConfigureAccessPointRequest
+//
+// General notes:
+//
+// Despite all fields being implicitly optional in proto3, all fields are
+// required to be specified unless their definition is explicitly marked with
+// the 'optional' proto keyword.
+//
+
+enum WifiAccessPointApplyConfigurationStatus
+{
+    WifiAccessPointApplyConfigurationStatusUnknown = 0;
+    WifiAccessPointApplyConfigurationStatusSucceeded = 1;
+    WifiAccessPointApplyConfigurationStatusAccessPointInvalid = 2;
+    WifiAccessPointApplyConfigurationStatusAccessPointNotEnabled = 3;
+    WifiAccessPointApplyConfigurationStatusBandInactive = 4;
+    WifiAccessPointApplyConfigurationStatusOperationNotSupported = 5;
+}
+
+message WifiAccessPointApplyConfigurationRequestConfigurationItem
+{
+    // An optional unique identifier for this configuration request.
+    optional string Id = 1;
+
+    // The band the AP should be configured for.
+    Microsoft.Net.Wifi.RadioBand Band = 2;
+
+    // Specifies the parts of the 'Configuration' field to be applied.
+    google.protobuf.FieldMask ConfigurationMask = 3;
+
+    // The configuration for the AP.
+    Microsoft.Net.Wifi.AccessPointConfiguration Configuration = 4;
+}
+
+message WifiAccessPointApplyConfigurationResultConfigurationItem
+{
+    // The unique identifier of the configuration that was requested to be
+    // applied. Not populated if not specified in the corresponding request.
+    optional string Id = 1;
+
+    // The band the AP was requested to be configured for.
+    Microsoft.Net.Wifi.RadioBand Band = 2;
+
+    // The status of applying the configuration.
+    WifiAccessPointApplyConfigurationStatus Status = 3;
+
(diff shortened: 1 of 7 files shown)


corbin-phipps:
Thanks for drafting up this API design proposal! After reviewing it, I'm starting to lean more toward having distinct APIs for changing individual configuration options. Although this current proposal allows clients to make fewer API calls, I feel like it's slightly more confusing due to the need for certain protobuf-specific things like `FieldMask`. Also, from an implementation standpoint, it seems like it would be easier to add new configuration options if we kept them all as distinct APIs

abeltrano:
> Thanks for drafting up this API design proposal! After reviewing it, I'm starting to lean more toward having distinct APIs for changing individual configuration options. Although this current proposal allows clients to make fewer API calls, I feel like it's slightly more confusing due to the need for certain protobuf-specific things like `FieldMask`. Also, from an implementation standpoint, it seems like it would be easier to add new configuration options if we kept them all as distinct APIs

Thanks. I sort of agree, this is not natural to use despite being a fairly common pattern, albeit, more in REST protocols. For alternatives, any comments on the other rough proposal in the `Reviewer Focus` section? Or, any other ideas/proposals?

corbin-phipps:
> > Thanks for drafting up this API design proposal! After reviewing it, I'm starting to lean more toward having distinct APIs for changing individual configuration options. Although this current proposal allows clients to make fewer API calls, I feel like it's slightly more confusing due to the need for certain protobuf-specific things like `FieldMask`. Also, from an implementation standpoint, it seems like it would be easier to add new configuration options if we kept them all as distinct APIs
> 
> Thanks. I sort of agree, this is not natural to use despite being a fairly common pattern, albeit, more in REST protocols. For alternatives, any comments on the other rough proposal in the `Reviewer Focus` section? Or, any other ideas/proposals?

I think the idea of having a core set of configuration options when enabling an AP could work and may be simpler than this current proposal. If we went the other route, I think we should still create the AP with whatever is absolutely necessary and unchanging, then have distinct APIs to set other fields that could be changed. For example, as far as I know, the `Dot11PhyType` of an AP isn't going to change, but the `RadioBand`, `Dot11AuthenticationAlgorithm`, and maybe more could potentially change after the AP is created. Does that make sense?

abeltrano:
> > > Thanks for drafting up this API design proposal! After reviewing it, I'm starting to lean more toward having distinct APIs for changing individual configuration options. Although this current proposal allows clients to make fewer API calls, I feel like it's slightly more confusing due to the need for certain protobuf-specific things like `FieldMask`. Also, from an implementation standpoint, it seems like it would be easier to add new configuration options if we kept them all as distinct APIs
> > 
> > 
> > Thanks. I sort of agree, this is not natural to use despite being a fairly common pattern, albeit, more in REST protocols. For alternatives, any comments on the other rough proposal in the `Reviewer Focus` section? Or, any other ideas/proposals?
> 
> I think the idea of having a core set of configuration options when enabling an AP could work and may be simpler than this current proposal. If we went the other route, I think we should still create the AP with whatever is absolutely necessary and unchanging, then have distinct APIs to set other fields that could be changed. For example, as far as I know, the `Dot11PhyType` of an AP isn't going to change, but the `RadioBand`, `Dot11AuthenticationAlgorithm`, and maybe more could potentially change after the AP is created. Does that make sense?

Yes, definitely. Though, the `Dot11PhyType` could be desirable to change to simulate APs that only support older PHYs like 802.11b. I'll push another draft PR (hopefully) today with this proposal.

protocol/protos/NetRemoteWifi.proto
```diff
@@ -3,37 +3,79 @@ syntax = "proto3";
...
+message WifiAccessPointApplyConfigurationRequestConfigurationItem
+{
+    // An optional unique identifier for this configuration request.
+    optional string Id = 1;
+
+    // The band the AP should be configured for.
+    Microsoft.Net.Wifi.RadioBand Band = 2;
+
+    // Specifies the parts of the 'Configuration' field to be applied.
+    google.protobuf.FieldMask ConfigurationMask = 3;
```
abeltrano, line 36: Thinking about this more, this might not be needed at all. In proto3, all fields are implicitly `optional`, so, the client only has to populate the fields they want to configure, and the server can just do a `item.has_<field>()` check to determine if it's present. If it is, then that field/property gets configured. The result structure would still need a `FieldMask` to indicate which fields failed configuration.

I don't know if this moves the needle any on this design or makes or more attractive, but I'm going to make this update for at least the sake of posterity.

abeltrano:
> I feel like it's slightly more confusing due to the need for certain protobuf-specific things like `FieldMask`. Also, from an implementation standpoint, it seems like it would be easier to add new configuration options if we kept them all as distinct APIs

FYI, I made an update removing the `FieldMask` from the request. It's still used in the result to indicate which things failed. Will still be sending another draft PR with the other alternative.

abeltrano:
Per review comments and some offline discussion, closing this PR in favor of the API proposed in #78.

PR #80, opened 2023-12-16
Add shared access point discovery mechanism

abeltrano:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

* Allow access points to be discovered, both statically via probes, and dynamically via presence change monitoring.

### Technical Details

* Add `AccessPointDiscoveryAgent` top-level class that is responsible for discovering access points.
* Re-implement `AccessPointManager` in terms of `AccessPointDiscoveryAgent` registration, hiding the access point creation logic from it, leaving construction to the OS-specific discovery agents with functionality specified in `IAccessPointDiscoveryAgentOperations`.
* Add `notstd::enable_make_protected` helper to aid safely instantiating classes that supported `shared_from_this()`.
* Export `notstd` library since it is used in exported libraries and is thus required as such.

### Test Results

* Newly added unit tests pass.

### Reviewer Focus

* None

### Future Work

* Additional unit tests exercising new `AccessPointManager` code are needed.
* Implement an `AccessPointDiscoveryAgent` on Linux.

### Checklist


src/common/wifi/apmanager/include/microsoft/net/wifi/AccessPointDiscoveryAgent.hxx
```diff
@@ -0,0 +1,124 @@
...
+    virtual void
+    StopImpl();
+
+    /**
+     * @brief Derived class implementation of asynchronous discovery probe.
+     *
+     * @return std::future<std::vector<std::shared_ptr<IAccessPoint>>>
+     */
+    virtual std::future<std::vector<std::shared_ptr<IAccessPoint>>>
+    ProbeAsyncImpl();
```
abeltrano, line 113: Consider factoring these into an interface like:

```C++
struct IAccessPointDiscoveryAgentOperations
{
    void Start() = 0;
    void Stop() = 0;
    std::future<std::vector<std::shared_ptr<IAccessPoint>>> ProbeAsync();
};
```

then taking a `std::unique_ptr<IAccessPointDiscoveryAgentOperations>` as a constructor argument to avoid polymorphism on `AccessPointDiscoveryAgent` directly.

PR #121, opened 2024-01-22
Add APIs to update AP configuration

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR adds three new APIs for updating the configuration of an AP after the AP is enabled: `WifiAccessPointSetPhyType`, `WifiAccessPointSetAuthenticationConfiguration`, and `WifiAccessPointSetFrequencyBands`. This allows netremote clients to update the PHY type, AKM suite(s), cipher suite(s), and frequency band(s) of an AP.

### Technical Details

* Added `WifiAccessPointSetAuthenticationConfiguration`, and `WifiAccessPointSetFrequencyBands` to `NetRemoteService.proto` and related types to `NetRemoteWifi.proto` and `WifiCore.proto`.
* Added placeholder API implementations to `NetRemoteService.hxx/cxx`.
* Added basic unit tests for all APIs.

### Test Results

Unit tests pass.

### Reviewer Focus

None.

### Future Work

* Need to complete the API implementations and improve associated unit tests.
* Update `Dot11AccessPointConfiguration` to use `Dot11AccessPointAuthenticationConfiguration` instead of a single `Dot11AuthenticationAlgorithm` and `Dot11CipherSuite`.

### Checklist


protocol/protos/NetRemoteWifi.proto
```diff
@@ -61,3 +61,39 @@ message WifiAccessPointDisableResult
...
+message WifiAccessPointSetPhyTypeResult
+{
+    string AccessPointId = 1;
+    WifiAccessPointOperationStatus Status = 2;
+}
+
+message WifiAccessPointSetAuthenticationMethodRequest
+{
+    string AccessPointId = 1;
+    Microsoft.Net.Wifi.Dot11AuthenticationAlgorithm AuthenticationAlgorithm = 2;
```
abeltrano, line 80: I know this mirrors the value specified in `AccessPointConfiguration`, so you've definitely done the right thing here. 

Looking forward, I think we need to expand this since the authentication method will require some additional data to configure it to work. For example, anything with a PSK will require the PSK to be specified and all such required configuration should all be provided together so the auth method can be atomically set from the client's perspective.

I'm not exactly sure how we would encode this, but at the very least, we probably need to wrap that configuration in separate messages/structures so they can evolve if necessary without needing to work hard to avoid breaking changes. 

Something like this could be one possible approach (just spitballing here):

```proto
message WifiAuthenticationMethodConfigurationPskPersonal
{
    Microsoft.Net.Wifi.SharedKey Psk;
    // ... whatever else is needed
}
message WifiAuthenticationMethodConfigurationEnterprise
{
    string RadiusServerAddress;
    // ... whatever else is needed
}
message WifiAuthenticationMethodConfiguration
{
   Microsoft.Net.Wifi.Dot11AuthenticationAlgorithm AuthenticationAlgorithm;
   oneof Configuration
   {
        WifiAuthenticationMethodConfigurationPskPersonal ConfigurationPskPersonal;
        WifiAuthenticationMethodConfigurationEnterprise ConfigurationEnterprise;
        // other auth method configuration structures ..
   }
   // ... whatever else is needed; other things can be safely added later without breaking backward compatibility.
}
```

We need to look at hostapd and see what it needs for each auth method, which can help guide how to encode that configuration here. I say "guide" instead of "define" because we shouldn't design only with hostapd in mind since that is a specific interpretation of how an AP is controlled. Anyway, check out this [hostapd.conf example file](https://w1.fi/cgit/hostap/tree/hostapd/hostapd.conf?h=hostap_2_10) which describes what each of the settings means. You can start with looking at the config args in this order of importance: `hw_mode`, `wpa`, `wpa_key_mgmt`, `wpa_pairwise`, `rsn_pairwise`, `group_cipher`, `group_mgmt_cipher`, `wpa_psk`, `wpa_passphrase`, `sae_password`, `wpa_psk_radius`. We would figure out which of these we need to set, then set it with the value(s) specified through the API using `Hostapd::SetProperty('config-property-name', 'property-value')`.

abeltrano (commented):
This is very good, mathcing the existing APIs we have in place to configure the AP. I left a few comments of things to explore a bit further for looking forward.

protocol/protos/NetRemoteWifi.proto
```diff
@@ -61,3 +61,39 @@ message WifiAccessPointDisableResult
...
+message WifiAccessPointSetPhyTypeRequest
+{
+    string AccessPointId = 1;
+    Microsoft.Net.Wifi.Dot11PhyType PhyType = 2;
+}
+
+message WifiAccessPointSetPhyTypeResult
+{
+    string AccessPointId = 1;
+    WifiAccessPointOperationStatus Status = 2;
```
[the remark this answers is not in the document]
corbin-phipps: I think setting `WifiAccessPointOperationStatusCodeOperationNotSupported` is good enough, especially since we can set a status message alongside the status code. That should be enough information for the client to go off of in my opinion.

protocol/protos/NetRemoteWifi.proto
corbin-phipps: Sounds good. I admit I didn't put too much thought into what was needed for the auth, I simply included the `Dot11AuthenticationAlgorithm` field, but it makes sense that certain auth methods need more information. I will take a look at hostapd and expand the API accordingly (the approach you gave looks pretty solid at first glance).

protocol/protos/WifiCore.proto
```diff
@@ -131,3 +131,31 @@ enum AccessPointState
...
+
+message WifiAuthenticationMethodConfiguration
+{
+    Microsoft.Net.Wifi.Dot11AuthenticationAlgorithm AuthenticationAlgorithm = 1;
+    oneof Configuration
+    {
+        WifiAuthenticationMethodConfigurationNone ConfigurationNone = 2;
+        WifiAuthenticationMethodConfigurationSharedKey ConfigurationSharedKey = 3;
+        WifiAuthenticationMethodConfigurationEnterprise ConfigurationEnterprise = 4;
+    }
```
corbin-phipps, line 160: If I'm not mistaken, these `Configuration` values should cover everything needed for all `Dot11AuthenticationAlgorithm` values:

`DOT11_AUTH_ALGO_80211_OPEN`: `ConfigurationNone`
`DOT11_AUTH_ALGO_80211_SHARED_KEY`: `ConfigurationSharedKey`
`DOT11_AUTH_ALGO_WPA`: `ConfigurationEnterprise`
`DOT11_AUTH_ALGO_WPA_PSK`: `ConfigurationSharedKey`
`DOT11_AUTH_ALGO_WPA_NONE`: This is not supported
`DOT11_AUTH_ALGO_RSNA`: `ConfigurationEnterprise`
`DOT11_AUTH_ALGO_RSNA_PSK`: `ConfigurationSharedKey`
`DOT11_AUTH_ALGO_WPA3`: `ConfigurationEnterprise`
`DOT11_AUTH_ALGO_WPA3_ENT_192`: `ConfigurationEnterprise`
`DOT11_AUTH_ALGO_WPA3_SAE`: `ConfigurationSharedKey`
`DOT11_AUTH_ALGO_OWE`: `ConfigurationNone`
`DOT11_AUTH_ALGO_WPA3_ENT`: `ConfigurationEnterprise`

protocol/protos/NetRemoteService.proto
```diff
@@ -11,4 +11,7 @@ service NetRemote
     rpc WifiEnumerateAccessPoints (Microsoft.Net.Remote.Wifi.WifiEnumerateAccessPointsRequest) returns (Microsoft.Net.Remote.Wifi.WifiEnumerateAccessPointsResult);
     rpc WifiAccessPointEnable (Microsoft.Net.Remote.Wifi.WifiAccessPointEnableRequest) returns (Microsoft.Net.Remote.Wifi.WifiAccessPointEnableResult);
     rpc WifiAccessPointDisable (Microsoft.Net.Remote.Wifi.WifiAccessPointDisableRequest) returns (Microsoft.Net.Remote.Wifi.WifiAccessPointDisableResult);
+    rpc WifiAccessPointSetPhyType (Microsoft.Net.Remote.Wifi.WifiAccessPointSetPhyTypeRequest) returns (Microsoft.Net.Remote.Wifi.WifiAccessPointSetPhyTypeResult);
+    rpc WifiAccessPointSetAuthenticationMethod (Microsoft.Net.Remote.Wifi.WifiAccessPointSetAuthenticationMethodRequest) returns (Microsoft.Net.Remote.Wifi.WifiAccessPointSetAuthenticationMethodResult);
+    rpc WifiAccessPointSetEncryptionMethod (Microsoft.Net.Remote.Wifi.WifiAccessPointSetEncryptionMethodRequest) returns (Microsoft.Net.Remote.Wifi.WifiAccessPointSetEncryptionMethodResult);
```
abeltrano, line 16: I'm wondering if we should consider consolidating these two. While each one has its own configuration value (in hostapd, and likely elsewhere), they are more or less linked. Since all fields are implicitly optionally in proto3, the client will have the option to specify either the authentication algorithm suite(s), the cipher suite(s), or both. Eg.

```proto
enum Dot11WepKeyType 
{
    WepKeyType40,
    WepKeyType140,
}
message Dot11WepKey
{
    Dot11WepKeyType Type;
    bytes Key;
}
message Dot11CipherSuiteConfiguration
{
    Dot11WepKey WepKey;
     // ... other cipher info for specific cipher algorithms     
}
message WifiAccessPointAuthenticationConfiguration
{
    // below map implies removing AuthenticationAlgorithm field from WifiAuthenticationMethodConfiguration message
    map<Dot11AuthenticationAlgorithm, WifiAuthenticationMethodConfiguration> Dot11AuthenticationAndKeyManagement AkmSuites;
    map<Dot11CipherAlgorithm, Dot11CipherSuiteConfiguration> CipherSuites;
}
```

protocol/protos/NetRemoteWifi.proto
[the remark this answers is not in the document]
abeltrano: Resolving.

protocol/protos/NetRemoteService.proto
corbin-phipps: It makes to consolidate them. In Windows, there is a `DOT11_AUTH_CIPHER_PAIR` structure that does exactly that.

Not sure if I agree with using a map though since the API is for setting the new configuration for the AP, which would just be one key-value pair of "auth algo"-"auth config". Is there a reason to set multiple?

corbin-phipps: I made some changes to combine the APIs but I did it slightly differently than what you proposed - please let me know what you think. I also didn't add WEP types yet since I'm not sure if the keys should be part of the auth method configuration or the encryption method configuration.

abeltrano: > Not sure if I agree with using a map though since the API is for setting the new configuration for the AP, which would just be one key-value pair of "auth algo"-"auth config". Is there a reason to set multiple?

Multiple AKMs can be supported on the same radio (eg. WPA-Personal and WPA-Enterprise). That's why programs like hostapd accept multiple (eg. see the `wpa_key_mgmt` setting).

I'm ok with accepting/supporting only a single AKM at any one time, but the design shouldn't preclude being able to accept multiple simultaneously in the future.

corbin-phipps: I ended up using `repeated` instead of `map`. It would make sense to have `map<Microsoft.Net.Wifi.Dot11AkmSuite, Microsoft.Net.Wifi.Dot11AkmSuiteConfiguration>`, but `Dot11AkmSuite` is an enum which can't be used as a map key and it seems silly to make a wrapper for that

abeltrano: Good point about not be able to use an enum as a map key (the workaround is ugly).

abeltrano:
> ### Test Results
> Unit tests currently fail due to failures in the API implementations when creating an access point controller.

This is probably because the unit tests are using fake access point ids, and the server's `AccessPointManager` doesn't know about any APs with those ids. Probably we need to create software APs using the mac80211_hwsim driver which will be discovered by the server, like is done for the `wpa-controller` tests. Unfortunately, that again necessitates running the tests as root. Alternatively, we might be able to provide a fake ap discovery agent that serves up mocked APs.

corbin-phipps:
> > ### Test Results
> > Unit tests currently fail due to failures in the API implementations when creating an access point controller.
> 
> This is probably because the unit tests are using fake access point ids, and the server's `AccessPointManager` doesn't know about any APs with those ids. Probably we need to create software APs using the mac80211_hwsim driver which will be discovered by the server, like is done for the `wpa-controller` tests. Unfortunately, that again necessitates running the tests as root. Alternatively, we might be able to provide a fake ap discovery agent that serves up mocked APs.

Yes, you're absolutely right. I removed some `REQUIRE` statements from the unit tests so that they would pass as is until we added a way to create APs for the tests. It would be nice to avoid mocked APs if possible, but that's an option if needed

abeltrano:
> > > ### Test Results
> > > Unit tests currently fail due to failures in the API implementations when creating an access point controller.
> > 
> > 
> > This is probably because the unit tests are using fake access point ids, and the server's `AccessPointManager` doesn't know about any APs with those ids. Probably we need to create software APs using the mac80211_hwsim driver which will be discovered by the server, like is done for the `wpa-controller` tests. Unfortunately, that again necessitates running the tests as root. Alternatively, we might be able to provide a fake ap discovery agent that serves up mocked APs.
> 
> Yes, you're absolutely right. I removed some `REQUIRE` statements from the unit tests so that they would pass as is until we added a way to create APs for the tests. It would be nice to avoid mocked APs if possible, but that's an option if needed

Looking at this now, the `AccessPointManager` part will be easier than expected as it has `AddAccessPoint` and `RemoveAccessPoint` functions; they're private, but I don't see much harm in making them public, allowing the test to use them and add arbitrary access points.

We can trivially extend the existing [`AccessPointTest`](https://github.com/microsoft/netremote/blob/develop/tests/unit/wifi/helpers/include/microsoft/net/wifi/test/AccessPointTest.hxx), `AccessPointFactoryTest`, and [`AccessPointControllerTest`](https://github.com/microsoft/netremote/blob/develop/tests/unit/wifi/helpers/include/microsoft/net/wifi/test/AccessPointControllerTest.hxx) with the functionality needed in the tests. For example, `AccessPointTest` would just save whatever values requested to be set in members. Should be straight-forward. I'll take a look at doing this later today.

abeltrano:
> Looking at this now, the `AccessPointManager` part will be easier than expected as it has `AddAccessPoint` and `RemoveAccessPoint` functions; they're private, but I don't see much harm in making them public, allowing the test to use them and add arbitrary access points. Alternatively, a thin wrapper class exposing the `Add/Remove` functions is another option that preserves the existing function visibility.
> 
> We can trivially extend the existing [`AccessPointTest`](https://github.com/microsoft/netremote/blob/develop/tests/unit/wifi/helpers/include/microsoft/net/wifi/test/AccessPointTest.hxx), `AccessPointFactoryTest`, and [`AccessPointControllerTest`](https://github.com/microsoft/netremote/blob/develop/tests/unit/wifi/helpers/include/microsoft/net/wifi/test/AccessPointControllerTest.hxx) with the functionality needed in the tests. For example, `AccessPointTest` would just save whatever values requested to be set in members. Should be straight-forward. I'll take a look at doing this later today.

Done in #136.

protocol/protos/WifiCore.proto
```diff
@@ -157,3 +157,44 @@ enum Dot11AccessPointState
...
+message Dot11AkmSuiteConfiguration
+{
+    Microsoft.Net.Wifi.Dot11AuthenticationAlgorithm AuthenticationAlgorithm = 1;
+    Microsoft.Net.Wifi.Dot11AkmSuite AkmSuite = 2;
+    oneof Configuration
+    {
+        Dot11AkmSuiteConfigurationNone ConfigurationNone = 3;
+        Dot11AkmSuiteConfigurationSharedKey ConfigurationSharedKey = 4;
+        Dot11AkmSuiteConfigurationEnterprise ConfigurationEnterprise = 5;
+    }
```
abeltrano, line 192: FYI, I'm still trying to parse out whether this is the best encoding of all the auth args w.r.t. to the 802.11 specification. I think it's very close, if not completely correct.

I'm giving the auth-related parts of the spec a careful read through to hopefully resolve this.

PR #139, opened 2024-01-27
Implement WifiAccessPointSetPhyType

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR adds a new netremote API, `WifiAccessPointSetPhyType`, to allow clients to update the PHY type in the configuration of an AP.

### Technical Details

* Added `WifiAccessPointSetPhyType` to `NetRemoteService.proto` and related types to `NetRemoteWifi.proto`.
* Implemented API in `NetRemoteService.*`.
* Added `SetProtocol` function to `IAccessPointController` interface and in its implementations (e.g. `AccessPointControllerLinux`.
* Added unit test.

### Test Results

Unit tests pass.

### Reviewer Focus

* `SetProtocol` sets the `hw_mode` hostapd property, as well as `wmm_enabled` and `ieee80211*`  properties where applicable. Are others needed, such as `ht_capab` for ieee80211n?

* In `IeeeProtocolToHostapdHwMode`, the conversion from `ieee80211n` to hostapd `hw_mode` can be either `a` or `g` depending on the band, currently it's set to `a`. How should we handle this? Is there a way to check the bands that are currently set in hostapd somewhere and choose the best option, or should we always set this to `a` and update it later if a client wants to set a specific frequency band?

### Future Work

None.

### Checklist

### DIFF
--- a/src/common/service/NetRemoteService.cxx
+++ b/src/common/service/NetRemoteService.cxx
@@ -494,89 +494,9 @@ NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerConte
     return grpc::Status::OK;
 }
 
-using Microsoft::Net::Wifi::Dot11AccessPointAuthenticationConfiguration;
 using Microsoft::Net::Wifi::Dot11AkmSuite;
-using Microsoft::Net::Wifi::Dot11AkmSuiteConfiguration;
 using Microsoft::Net::Wifi::Dot11AuthenticationAlgorithm;
 using Microsoft::Net::Wifi::Dot11CipherSuite;
-
-::grpc::Status
-NetRemoteService::WifiAccessPointSetAuthenticationConfiguration([[maybe_unused]] ::grpc::ServerContext* context, const ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetAuthenticationConfigurationRequest* request, ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetAuthenticationConfigurationResult* response)
-{
-    LOGD << std::format("Received WifiAccessPointSetAuthenticationConfiguration request for access point id {}", request->accesspointid());
-
-    WifiAccessPointOperationStatus status{};
-
-    const auto& authenticationConfiguration = request->authenticationconfiguration();
-    const auto& akmSuites = authenticationConfiguration.akmsuites();
-    const auto& cipherSuites = authenticationConfiguration.ciphersuites();
-
-    if (std::empty(akmSuites) && std::empty(cipherSuites)) {
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter);
-        status.set_message("No AKM suite or cipher suite provided");
-    } else if (!std::empty(akmSuites) && std::ranges::find(akmSuites, Dot11AkmSuite::Dot11AkmSuiteUnknown, &Dot11AkmSuiteConfiguration::akmsuite) != std::cend(akmSuites)) {
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter);
-        status.set_message("Invalid AKM suite provided");
-    } else if (!std::empty(akmSuites) && std::ranges::find(akmSuites, Dot11AuthenticationAlgorithm::Dot11AuthenticationAlgorithmUnknown, &Dot11AkmSuiteConfiguration::authenticationalgorithm) != std::cend(akmSuites)) {
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter);
-        status.set_message("Invalid authentication algorithm provided");
-    } else if (!std::empty(cipherSuites) && std::ranges::find(cipherSuites, Dot11CipherSuite::Dot11CipherSuiteUnknown) != std::cend(cipherSuites)) {
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter);
-        status.set_message("Invalid cipher suite provided");
-    } else {
-        auto accessPointWeak = m_accessPointManager->GetAccessPoint(request->accesspointid());
-        auto accessPointController = detail::IAccessPointWeakToAccessPointController(accessPointWeak);
-        if (!accessPointController) {
-            LOGE << std::format("Failed to create controller for access point {}", request->accesspointid());
-            status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeAccessPointInvalid);
-            status.set_message(std::format("Failed to create controller for access point {}", request->accesspointid()));
-        }
-
-        // TODO: Use accessPointController to set authentication configuration.
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeSucceeded);
-    }
-
-    response->set_accesspointid(request->accesspointid());
-    *response->mutable_status() = std::move(status);
-
-    return grpc::Status::OK;
-}
-
-using Microsoft::Net::Wifi::Dot11FrequencyBand;
-
-::grpc::Status
-NetRemoteService::WifiAccessPointSetFrequencyBands([[maybe_unused]] ::grpc::ServerContext* context, const ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetFrequencyBandsRequest* request, ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetFrequencyBandsResult* response)
-{
-    LOGD << std::format("Received WifiAccessPointSetFrequencyBands request for access point id {}", request->accesspointid());
-
-    WifiAccessPointOperationStatus status{};
-
-    const auto& frequencyBands = request->frequencybands();
-
-    if (std::empty(frequencyBands)) {
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter);
-        status.set_message("No frequency band provided");
-    } else if (std::ranges::find(frequencyBands, Dot11FrequencyBand::Dot11FrequencyBandUnknown) != std::cend(frequencyBands)) {
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter);
-        status.set_message("Invalid frequency band provided");
-    } else {
-        auto accessPointWeak = m_accessPointManager->GetAccessPoint(request->accesspointid());
-        auto accessPointController = detail::IAccessPointWeakToAccessPointController(accessPointWeak);
-        if (!accessPointController) {
-            LOGE << std::format("Failed to create controller for access point {}", request->accesspointid());
-            status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeAccessPointInvalid);
-            status.set_message(std::format("Failed to create controller for access point {}", request->accesspointid()));
-        }
-
-        // TODO: Use accessPointController to set frequency bands.
-        status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeSucceeded);
-    }
-
-    response->set_accesspointid(request->accesspointid());
-    *response->mutable_status() = std::move(status);
-
-    return grpc::Status::OK;
-}
 
 bool
 NetRemoteService::ValidateWifiAccessPointEnableRequest(const ::Microsoft::Net::Remote::Wifi::WifiAccessPointEnableRequest* request, ::Microsoft::Net::Remote::Wifi::WifiAccessPointOperationStatus& status)
--- a/src/common/service/NetRemoteService.cxx
+++ b/src/common/service/NetRemoteService.cxx
@@ -85,7 +85,6 @@ NetRemotePhyTypeToIeeeProtocol(Dot11PhyType phyType)
 
 using Microsoft::Net::Wifi::Dot11FrequencyBand;
 using Microsoft::Net::Wifi::Ieee80211FrequencyBand;
-using Microsoft::Net::Wifi::Dot11FrequencyBand;
 
 Dot11FrequencyBand
 IeeeDot11FrequencyBandToNetRemoteDot11FrequencyBand(Ieee80211FrequencyBand ieeeDot11FrequencyBand)
--- a/src/common/service/NetRemoteService.cxx
+++ b/src/common/service/NetRemoteService.cxx
@@ -440,7 +440,7 @@ NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerConte
     WifiAccessPointOperationStatus status{};
     status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeSucceeded);
 
-    auto handleFailure = [&](WifiAccessPointOperationStatusCode statusCode, const std::string& statusMessage) {
+    auto handleFailure = [&](WifiAccessPointOperationStatusCode statusCode, std::string statusMessage) {
         LOGE << statusMessage;
 
         status.set_code(statusCode);
@@ -454,14 +454,14 @@ NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerConte
 
     // Check if PHY type is provided.
     if (request->phytype() == Dot11PhyType::Dot11PhyTypeUnknown) {
-        handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter, "No PHY type provided");
+        return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter, "No PHY type provided");
     }
 
     // Create an AP controller for the requested AP.
     auto accessPointWeak = m_accessPointManager->GetAccessPoint(request->accesspointid());
     auto accessPointController = detail::IAccessPointWeakToAccessPointController(accessPointWeak);
     if (!accessPointController) {
-        handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeAccessPointInvalid, std::format("Failed to create controller for access point {}", request->accesspointid()));
+        return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeAccessPointInvalid, std::format("Failed to create controller for access point {}", request->accesspointid()));
     }
 
     // Convert PHY type to Ieee80211 protocol.
@@ -471,21 +471,21 @@ NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerConte
     try {
         const auto& supportedIeeeProtocols = accessPointController->GetCapabilities().Protocols;
         if (std::ranges::find(supportedIeeeProtocols, ieeeProtocol) == std::cend(supportedIeeeProtocols)) {
-            handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("PHY type not supported by access point {}", request->accesspointid()));
+            return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("PHY type not supported by access point {}", request->accesspointid()));
         }
     } catch (const AccessPointControllerException& apce) {
         LOGE << std::format("Failed to get capabilities for access point {} ({})", request->accesspointid(), apce.what());
-        handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("Failed to get capabilities for access point {}", request->accesspointid()));
+        return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("Failed to get capabilities for access point {}", request->accesspointid()));
     }
 
     // Set the Ieee80211 protocol.
     try {
-        if (!accessPointController->SetIeeeProtocol(ieeeProtocol)) {
-            handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("Failed to set PHY type for access point {}", request->accesspointid()));
+        if (!accessPointController->SetProtocol(ieeeProtocol)) {
+            return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("Failed to set PHY type for access point {}", request->accesspointid()));
         }
     } catch (const AccessPointControllerException& apce) {
         LOGE << std::format("Failed to set Ieee80211 protocol for access point {} ({})", request->accesspointid(), apce.what());
-        handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("Failed to set PHY type for access point {}", request->accesspointid()));
+        return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported, std::format("Failed to set PHY type for access point {}", request->accesspointid()));
     }
 
     response->set_accesspointid(request->accesspointid());
--- a/src/common/service/NetRemoteService.cxx
+++ b/src/common/service/NetRemoteService.cxx
@@ -464,9 +464,10 @@ NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerConte
         handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeAccessPointInvalid, std::format("Failed to create controller for access point {}", request->accesspointid()));
     }
 
+    // Convert PHY type to Ieee80211 protocol.
     auto ieeeProtocol = detail::NetRemotePhyTypeToIeeeProtocol(request->phytype());
 
-    // Check if PHY type is supported by AP.
+    // Check if Ieee80211 protocol is supported by AP.
     try {
         const auto& supportedIeeeProtocols = accessPointController->GetCapabilities().Protocols;
         if (std::ranges::find(supportedIeeeProtocols, ieeeProtocol) == std::cend(supportedIeeeProtocols)) {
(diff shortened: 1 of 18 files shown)


src/common/wifi/core/include/microsoft/net/wifi/IAccessPointController.hxx
```diff
@@ -60,6 +61,17 @@ struct IAccessPointController
...
+    /**
+     * @brief Set the PHY type of the access point.
+     *
+     * @param phyType The PHY type to be set
+     * @return true
+     * @return false
+     *
+     */
+    virtual bool
+    SetPhyType(Dot11PhyType phyType) = 0;
```
abeltrano, line 74: See comments in `AccessPointControllerLinux.[hxx|cxx]` regarding layering dependencies and use of `Ieee80211*` types instead of the protobuf types in this layer.

src/common/wifi/core/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ target_sources(wifi-core
 target_link_libraries(wifi-core
     PUBLIC
         notstd
+        ${PROJECT_NAME}-protocol
```
abeltrano, line 28: Need to remove this to eliminate upward dependency, see related comments elsewhere.

src/common/service/NetRemoteService.cxx
```diff
@@ -316,6 +316,19 @@ IAccessPointWeakToNetRemoteAccessPointResultItem(std::weak_ptr<Microsoft::Net::W
...
 }
 
+std::unique_ptr<Microsoft::Net::Wifi::IAccessPointController>
+IAccessPointWeakToAccessPointController(std::weak_ptr<Microsoft::Net::Wifi::IAccessPoint>& accessPointWeak)
+{
+    auto accessPoint = accessPointWeak.lock();
+    if (accessPoint != nullptr) {
+        return accessPoint->CreateController();
+    } else {
+        LOGE << "Failed to retrieve access point";
```
abeltrano, line 326: Adding why it could not be retrieved would make decoding what happened from the logs a bit easier:

```suggestion
        LOGE << "Failed to retrieve access point as it is no longer valid";
```

src/linux/wifi/core/CMakeLists.txt
```diff
@@ -24,4 +24,5 @@ target_link_libraries(wifi-core-linux
         libnl-helpers
         wifi-core
         wpa-controller
+        ${PROJECT_NAME}-protocol
```
abeltrano, line 27: Remove (see other comments).

src/common/service/NetRemoteService.cxx
```diff
@@ -389,6 +399,125 @@ NetRemoteService::WifiAccessPointDisable([[maybe_unused]] ::grpc::ServerContext*
...
+    } else {
+        auto accessPointWeak = m_accessPointManager->GetAccessPoint(request->accesspointid());
+        auto accessPointController = detail::IAccessPointWeakToAccessPointController(accessPointWeak);
+        if (!accessPointController) {
+            LOGE << std::format("Failed to create controller for access point {}", request->accesspointid());
+            status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeAccessPointInvalid);
+            status.set_message(std::format("Failed to create controller for access point {}", request->accesspointid()));
+        }
+
+        if (!accessPointController->SetPhyType(request->phytype())) {
```
abeltrano, line 424: One thing (I think) we need to account for is whether the specified phy type is supported, where in the stack to perform this check, and how to signal any errors to the API client related to arguments that aren't supported. Currently IAccessPointController::SetPhyType() returns a `bool` which I think is fine, but that means the check for support will need to be done at a higher level since the `bool` return type can't provide any detailed information beyond success/fail.

This information is in the `Ieee80211AccessPointCapabilities` object, which can be obtained using `IAccessPointController::GetCapabilities()`, so the API implementation could check this directly.

I suppose an alternative is to not do the check at all, and just let the lowest layer fail on bad/invalid arguments. My only concern is that the caller then can't differentiate between the cases where the failure occurred due to some temporary problem, or whether the arguments were truly unsupported. That's why I lean toward adding an explicit check for support against the requested phy type.

This pattern will apply to other APIs that are setting/changing configuration values, so we should think deeply about how best to handle this and apply the solution to all of those APIs.

abeltrano (changes requested):

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -119,6 +166,20 @@ AccessPointControllerLinux::GetIsEnabled()
...
+    Wpa::HostapdHwMode hwMode = detail::Dot11PhyTypeToHostapdHwMode(phyType);
+    const auto& propertyName = Wpa::ProtocolHostapd::PropertyNameSetHwMode;
+    const auto& propertyValue = detail::HostapdHwModeToPropertyValue(hwMode);
+
+    try {
+        return m_hostapd.SetProperty(propertyName, propertyValue);
+    } catch (const Wpa::HostapdException& ex) {
+        throw AccessPointControllerException(std::format("Failed to set PHY type for interface {} ({})", GetInterfaceName(), ex.what()));
+    }
+}
```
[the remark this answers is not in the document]
corbin-phipps: Thanks, I'll need to add onto this a bit more because some PHY types need to set more than just `hw_mode`, such as `ieee80211n=1` or `ieee80211ac=1` for n/ac PHY types

src/common/service/NetRemoteService.cxx
corbin-phipps: Good ideas. In general, I am a fan of identifying issues as early as possible to more easily identify when the issue occurs. So, I'm leaning toward checking `Ieee80211AccessPointCapabilities` in the API implementation directly (`NetRemoteService::WifiAccessPointSetPhyType`) so that an informative status message can explain that the PHY type is not supported. Then if `AccessPointControllerLinux::SetPhyType` still returns `false`, then we know there was an issue with actually setting the PHY type value

abeltrano: Sounds great.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -74,6 +75,52 @@ Nl80211WiphyToIeee80211Protocols(const Nl80211Wiphy& nl80211Wiphy)
...
+    case Wpa::HostapdHwMode::Ieee80211a:
+        return Wpa::ProtocolHostapd::PropertySetHwModeValueA;
+    case Wpa::HostapdHwMode::Ieee80211ad:
+        return Wpa::ProtocolHostapd::PropertySetHwModeValueAD;
+    case Wpa::HostapdHwMode::Ieee80211any:
+        return Wpa::ProtocolHostapd::PropertySetHwModeValueAny;
+    default: // case Wpa::HostapdHwMode::Unknown
+        throw AccessPointControllerException(std::format("Invalid hw_mode value {}", magic_enum::enum_name(hwMode)));
+    }
+}
```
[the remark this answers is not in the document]
corbin-phipps: I changed the first one to `IeeeProtocolToHostapdHwMode` but kept it in this layer. This layer has the implementation for `SetIeeeProtocol(Ieee80211Protocol ieeeProtocol)` (formerly `SetPhyType`), so it seems like this is where the conversion to hostapd hw_mode should occur, but I will move it elsewhere if you disagree

src/common/wifi/core/include/microsoft/net/wifi/IAccessPointController.hxx
```diff
@@ -60,6 +61,17 @@ struct IAccessPointController
...
+    /**
+     * @brief Set the Ieee80211 protocol of the access point.
+     *
+     * @param ieeeProtocol The Ieee80211 protocol to be set
+     * @return true
+     * @return false
+     *
+     */
+    virtual bool
+    SetIeeeProtocol(Microsoft::Net::Wifi::Ieee80211Protocol ieeeProtocol) = 0;
```
abeltrano, line 74: The protocol is assumed to be IEEE 802.11 at this layer, so we should drop the `Ieee` part from the name:

```suggestion
    SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol ieeeProtocol) = 0;
```

There should be no case for ambiguity since the API layer isn't involved here.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -74,6 +75,50 @@ Nl80211WiphyToIeee80211Protocols(const Nl80211Wiphy& nl80211Wiphy)
...
+    case Wpa::HostapdHwMode::Ieee80211g:
+        return Wpa::ProtocolHostapd::PropertySetHwModeValueG;
+    case Wpa::HostapdHwMode::Ieee80211a:
+        return Wpa::ProtocolHostapd::PropertySetHwModeValueA;
+    case Wpa::HostapdHwMode::Ieee80211ad:
+        return Wpa::ProtocolHostapd::PropertySetHwModeValueAD;
+    case Wpa::HostapdHwMode::Ieee80211any:
+        return Wpa::ProtocolHostapd::PropertySetHwModeValueAny;
+    default: // case Wpa::HostapdHwMode::Unknown
+        throw AccessPointControllerException(std::format("Invalid hw_mode value {}", magic_enum::enum_name(hwMode)));
```
abeltrano, line 119: nit: Include "hostapd" in the error message for more clarity.

```suggestion
        throw AccessPointControllerException(std::format("Invalid hostapd hw_mode value {}", magic_enum::enum_name(hwMode)));
```

src/common/service/NetRemoteService.cxx
```diff
@@ -389,6 +424,156 @@ NetRemoteService::WifiAccessPointDisable([[maybe_unused]] ::grpc::ServerContext*
...
+NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerContext* context, const ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetPhyTypeRequest* request, ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetPhyTypeResult* response)
+{
+    using Microsoft::Net::Wifi::Ieee80211AccessPointCapabilities;
+
+    LOGD << std::format("Received WifiAccessPointSetPhyType request for access point id {}", request->accesspointid());
+
+    WifiAccessPointOperationStatus status{};
+    status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeSucceeded);
+
+    auto handleFailure = [&](WifiAccessPointOperationStatusCode statusCode, const std::string& statusMessage) {
```
abeltrano, line 439: nit: It makes sense to generally prefer `std::string_view` over `const std::string&` when a non-owning, immutable copy is being passed (eg. read-only scenarios) as `std::string_view` is lighter weight than the reference, and conveys non-ownership by definition. 

Here a copy is being made, so it probably makes even more sense to just use `std::string` instead, then the caller is 100% aware that a copy is being made and this functions owns that copy and can do whatever it wants with it. 

```suggestion
    auto handleFailure = [&](WifiAccessPointOperationStatusCode statusCode, std::string statusMessage) {
```

src/common/service/NetRemoteService.cxx
```diff
@@ -389,6 +424,156 @@ NetRemoteService::WifiAccessPointDisable([[maybe_unused]] ::grpc::ServerContext*
...
+
+        response->set_accesspointid(request->accesspointid());
+        *response->mutable_status() = std::move(status);
+
+        return grpc::Status::OK;
+    };
+
+    // Check if PHY type is provided.
+    if (request->phytype() == Dot11PhyType::Dot11PhyTypeUnknown) {
+        handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter, "No PHY type provided");
```
abeltrano, line 453: Do you mean to `return` here? It seems like it since the lambda is returning a `grpc::Status`. Same applies to other instances where `handleFailure` is used in this function.

```suggestion
        return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter, "No PHY type provided");
```

abeltrano (changes requested):

src/linux/wifi/core/AccessPointControllerLinux.cxx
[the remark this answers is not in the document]
abeltrano: Yep, that makes sense. I was wrong to include the 2nd one since that doesn't involve API/protobuf types 👍

src/common/service/NetRemoteService.cxx
corbin-phipps: Yes you're right, there should be `return`s at each call to `handleFailure`

src/common/service/NetRemoteService.cxx
corbin-phipps: Oh never mind, it seems like `status.set_message` makes a copy

abeltrano:
I re-targeted this to [`user/corbinphipps/add-update-config-apis`](https://github.com/microsoft/netremote/tree/user/corbinphipps/add-update-config-apis) so it's easier to see the changes compared to there since it seems this is a child branch of that.

abeltrano:
> I re-targeted this to [`user/corbinphipps/add-update-config-apis`](https://github.com/microsoft/netremote/tree/user/corbinphipps/add-update-config-apis) so it's easier to see the changes compared to there since it seems this is a child branch of that.

Nevermind, I now re-targeted it such that it can be applied directly to `develop` without the changes in `user/corbinphipps/add-update-config-apis`. Limiting the scope should make it easier to review and test. Overall, it probably makes sense to complete this PR first.

src/common/service/NetRemoteService.cxx
```diff
@@ -389,6 +427,76 @@ NetRemoteService::WifiAccessPointDisable([[maybe_unused]] ::grpc::ServerContext*
...
+    if (!accessPointController) {
+        return handleFailure(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeAccessPointInvalid, std::format("Failed to create controller for access point {}", request->accesspointid()));
+    }
+
+    // Convert PHY type to Ieee80211 protocol.
+    auto ieeeProtocol = detail::NetRemotePhyTypeToIeeeProtocol(request->phytype());
+
+    // Check if Ieee80211 protocol is supported by AP.
+    try {
+        const auto& supportedIeeeProtocols = accessPointController->GetCapabilities().Protocols;
```
abeltrano, line 471: This is a dangling reference since the capabilities object doesn't exist after this line completes. So, need to update this to either make a copy of the protocols, or, just keep the capabilities as a local. I'd probably do the latter so we could inspect the full capabilities if debugging.
```suggestion
        auto accessPointCapabilities = accessPointController->GetCapabilities();
        const auto& supportedIeeeProtocols = accessPointCapabilities.Protocols;
```

src/common/service/NetRemoteService.cxx
```diff
@@ -389,6 +427,76 @@ NetRemoteService::WifiAccessPointDisable([[maybe_unused]] ::grpc::ServerContext*
...
+
+::grpc::Status
+NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerContext* context, const ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetPhyTypeRequest* request, ::Microsoft::Net::Remote::Wifi::WifiAccessPointSetPhyTypeResult* response)
+{
+    using Microsoft::Net::Wifi::Ieee80211AccessPointCapabilities;
+
+    LOGD << std::format("Received WifiAccessPointSetPhyType request for access point id {}", request->accesspointid());
+
+    WifiAccessPointOperationStatus status{};
+    status.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeSucceeded);
```
abeltrano, line 440: We should move this to the end of the function to avoid any possibility of it being set on an error, and just generally delay doing anything until it's needed.

src/common/service/NetRemoteService.cxx
```diff
@@ -389,6 +427,76 @@ NetRemoteService::WifiAccessPointDisable([[maybe_unused]] ::grpc::ServerContext*
...
+
+    response->set_accesspointid(request->accesspointid());
+    *response->mutable_status() = std::move(status);
+
+    return grpc::Status::OK;
+}
+
+using Microsoft::Net::Wifi::Dot11AkmSuite;
+using Microsoft::Net::Wifi::Dot11AuthenticationAlgorithm;
+using Microsoft::Net::Wifi::Dot11CipherSuite;
```
abeltrano, line 498: We can remove this from this PR, right?

```suggestion

```

src/common/wifi/core/include/microsoft/net/wifi/IAccessPointController.hxx
```diff
@@ -60,6 +61,17 @@ struct IAccessPointController
      */
     virtual Ieee80211AccessPointCapabilities
     GetCapabilities() = 0;
+
+    /**
+     * @brief Set the Ieee80211 protocol of the access point.
+     *
+     * @param ieeeProtocol The Ieee80211 protocol to be set
```
abeltrano, line 68: nit: missing period at end
```suggestion
     * @param ieeeProtocol The Ieee80211 protocol to be set.
```

src/linux/wifi/core/include/microsoft/net/wifi/AccessPointControllerLinux.hxx
```diff
@@ -43,6 +44,16 @@ struct AccessPointControllerLinux :
     virtual Ieee80211AccessPointCapabilities
     GetCapabilities() override;
 
+    /**
+     * @brief Set the Ieee80211 protocol
+     *
+     * @param ieeeProtocol The Ieee80211 protocol to be set
```
abeltrano, line 50: nit: style (missing periods)
```suggestion
     * @brief Set the Ieee80211 protocol.
     *
     * @param ieeeProtocol The Ieee80211 protocol to be set.
```

tests/unit/TestNetRemoteServiceClient.cxx
```diff
@@ -114,3 +115,74 @@ TEST_CASE("WifiAccessPointEnable API", "[basic][rpc][client][remote]")
...
+    using namespace Microsoft::Net::Remote::Wifi;
+    using namespace Microsoft::Net::Wifi;
+    using namespace Microsoft::Net::Wifi::Test;
+
+    constexpr auto SsidName{ "TestWifiAccessPointSetPhyType" };
+    constexpr auto InterfaceName{ "TestWifiAccessPointSetPhyType" };
+
+    auto apManagerTest = std::make_shared<AccessPointManagerTest>();
+    Ieee80211AccessPointCapabilities apCapabilities{
+        .Protocols{ Ieee80211Protocol::B, Ieee80211Protocol::G, Ieee80211Protocol::N, Ieee80211Protocol::A, Ieee80211Protocol::AC, Ieee80211Protocol::AD, Ieee80211Protocol::AX, Ieee80211Protocol::AX }
```
abeltrano, line 132: This has `AX` twice. Was `BE` intended, instead?

```suggestion
        .Protocols{ Ieee80211Protocol::B, Ieee80211Protocol::G, Ieee80211Protocol::N, Ieee80211Protocol::A, Ieee80211Protocol::AC, Ieee80211Protocol::AD, Ieee80211Protocol::AX, Ieee80211Protocol::BE }
```

Alternatively, if this is meant to be all of the protocols, we should define it as a constant (`constexpr`) at the top of the test file for all tests to use, and `magic_enum` can help with this:

```C++
namespace Microsoft::Net::Remote::Test
{
constexpr auto AllProtocols = magic_enum::enum_values<Microsoft::Net::Wifi::Ieee80211Protocol>();
} // Microsoft::Net::Remote::Test
```

tests/unit/wifi/helpers/AccessPointControllerTest.cxx
```diff
@@ -41,6 +41,17 @@ AccessPointControllerTest::GetCapabilities()
...
 }
 
+bool
+AccessPointControllerTest::SetProtocol([[maybe_unused]] Microsoft::Net::Wifi::Ieee80211Protocol ieeeProtocol)
+{
+    if (AccessPoint == nullptr) {
+        throw std::runtime_error("AccessPointControllerTest::SetIeeeProtocol called with null AccessPoint");
+    }
+
+    // TODO: Set Ieee80211 protocol
```
abeltrano, line 51: For this, we should add a `Ieee80211Protocol Protocol` public member to `AccessPointTest`, then set it here like this:

```suggestion
    AccessPoint->Protocol = ieeeProtocol;
```

abeltrano (changes requested):

tests/unit/TestNetRemoteServiceClient.cxx
abeltrano: Just as a word of caution, `magic_enum::enum_values()` (and other magic_enum "magic") will only work with enums that have values <= 255. We can extend it to `UINT16_MAX - 1` if necessary (it will slow down the build, so we don't do that atm since it's not needed).

It's not a concern for this instance, but just want to give you a heads-up in case you decide to use such functionality later. Some examples where this doesn't work are the `Ieee80211AkmSuite` and `Ieee80211CipherSuite` enums since they have an underlying type and values of type `uint32_t`.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -119,6 +164,35 @@ AccessPointControllerLinux::GetIsEnabled()
...
+        // Additively set other hostapd properties based on the protocol.
+        if (ieeeProtocol == Ieee80211Protocol::N || ieeeProtocol == Ieee80211Protocol::AC || ieeeProtocol == Ieee80211Protocol::AX) {
+            isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameWmmEnabled, Wpa::ProtocolHostapd::PropertyEnabled);
+            isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameIeee80211N, Wpa::ProtocolHostapd::PropertyEnabled);
+        }
+        if (ieeeProtocol == Ieee80211Protocol::AC || ieeeProtocol == Ieee80211Protocol::AX) {
+            isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameIeee80211AC, Wpa::ProtocolHostapd::PropertyEnabled);
+        }
+        if (ieeeProtocol == Ieee80211Protocol::AX) {
+            isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameIeee80211AX, Wpa::ProtocolHostapd::PropertyEnabled);
```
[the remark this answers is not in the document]
corbin-phipps: You're totally right, I had refactored this and messed up the return logic. Will fix

src/linux/wpa-controller/include/Wpa/ProtocolHostapd.hxx
```diff
@@ -184,12 +184,30 @@ struct ProtocolHostapd :
...
+    static constexpr auto PropertyHwModeValueAD = "ad";
+    static constexpr auto PropertyHwModeValueAny = "any";
+
+    static constexpr auto PropertyNameIeee80211N = "ieee80211n";
+    static constexpr auto PropertyNameDisable11N = "disable_11n";
+    static constexpr auto PropertyNameIeee80211AC = "ieee80211ac";
+    static constexpr auto PropertyNameDisable11AC = "disable_11ac";
+    static constexpr auto PropertyNameIeee80211AX = "ieee80211ax";
+    static constexpr auto PropertyNameDisable11AX = "disable_11ax";
+    static constexpr auto PropertyNameWmmEnabled = "wmm_enabled";
```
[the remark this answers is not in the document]
corbin-phipps: I didn't see that in the hostapd.conf example file, but there is `wmm_enabled` which is supposed to be set if you set `ieee80211n` or `ieee80211ac`

abeltrano:
> * `SetProtocol` sets the `hw_mode` hostapd property, as well as `wmm_enabled` and `ieee80211*`  properties where applicable. Are others needed, such as `ht_capab` for ieee80211n?

We'll need to figure this out eventually, but we can defer that and do it later on an as-needed, case-by-case basis. I'll create some issues to track that work. The only thing to verify now is whether or not hostapd will refuse to apply the `hw_mode` setting if the associated settings like `ht_capab`, `vht_capab`,  are not specified. As long as it doesn't, 

> * `AccessPointControllerTest::SetProtocol` currently just returns true. Is this okay to leave unimplemented since this is for a mocked AP?

This one is handled now right, so we can remove this part from the PR description?

> * In `IeeeProtocolToHostapdHwMode`, the conversion from `ieee80211n` to hostapd `hw_mode` can be either `a` or `g` depending on the band, currently it's set to `a`. How should we handle this? Is there a way to check the bands that are currently set in hostapd somewhere and choose the best option, or should we always set this to `a` and update it later if a client wants to set a specific frequency band?

This one is difficult because of the hostapd config file implementation. I think what you've done here is good for now, and we can plan to sort out the nuisances later.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -119,6 +164,38 @@ AccessPointControllerLinux::GetIsEnabled()
...
 
+bool
+AccessPointControllerLinux::SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol ieeeProtocol)
+{
+    bool isOk = false;
+    Wpa::HostapdHwMode hwMode = detail::IeeeProtocolToHostapdHwMode(ieeeProtocol);
+
+    try {
+        // Set the hostapd hw_mode property.
+        isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameHwMode, detail::HostapdHwModeToPropertyValue(hwMode));
```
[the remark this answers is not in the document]
corbin-phipps: I set `hw_mode` to `a` for `Ieee80211Protocol::BE`, but the hostapd.conf example file doesn't seem to mention anything about it so I'm not really sure what to put

src/linux/wpa-controller/include/Wpa/ProtocolHostapd.hxx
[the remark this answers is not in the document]
abeltrano: Are you synced to the `hostap_2_10` tag? Mine shows this:

image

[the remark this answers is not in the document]
corbin-phipps: Yes I am, but I still don't see anything regarding ieee80211be! So strange..

src/linux/wifi/core/AccessPointControllerLinux.cxx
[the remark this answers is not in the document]
abeltrano: We can keep the enum entry for `BE` since it's already there, but let's otherwise not handle `BE` support for now.

abeltrano (approved):
Ship it!

PR #144, opened 2024-02-01
Add API to update AP authentication configuration

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

Describe what the PR intends to achieve. If the change is a new feature, describe what it is. If the change is a bug fix, reference the issue being fixed. Provide any additional context and motivation for making the change. Eg.
> Simplify maintenance of the event loop. A well-known open source library can be used in place of the custom written code, making it easier to use, understand, and debug.

### Technical Details

Describe how the goals of the PR were achieved. Eg.
> Replace custom event loop code with libevent.

### Test Results

Describe which tests were carried out and a summary of their results. Eg.
> Verified Near Object session completed successfully with the following configurations:
>
> - [X] host as controlee
> - [X] host as controller

### Reviewer Focus

Describe what reviewers should focus on. Eg.

> The event loop could cause the service to exit early if it does not manage object lifetime correctly. Please pay careful attention to how this is achieved in the new event loop.

### Future Work

Describe any future work that is required as a result of this change. Eg.

>
> - Long-running stress testing needs to be completed.
> - The old event loop code needs to be removed once stress-testing with libevent has been completed.
>

### Checklist

- [ ] Build target `all` compiles cleanly.
- [ ] clang-format and clang-tidy deltas produced no new output.
- [ ] Newly added functions include doxygen-style comment block.


src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -119,6 +119,94 @@ HostapdHwModeToPropertyValue(Wpa::HostapdHwMode hwMode)
...
+    case Ieee80211AkmSuite::Tdls:
+        return "";
+    case Ieee80211AkmSuite::Sae:
+        return Wpa::ProtocolHostapd::PropertyWpaKeyMgmtValueSae;
+    case Ieee80211AkmSuite::FtSae:
+        return Wpa::ProtocolHostapd::PropertyWpaKeyMgmtValueFtSae;
+    case Ieee80211AkmSuite::ApPeerKey:
+        return "";
+    case Ieee80211AkmSuite::Ieee8021xSuiteB:
+        return "";
```
abeltrano, line 150: Need to add one for this:

```suggestion
    // Next line in ProtocolHostapd.hxx:
    static constexpr auto PropertyWpaKeyMgmtValueWpaEapSuiteB = "WPA-EAP-SUITE-B";

    case Ieee80211AkmSuite::Ieee8021xSuiteB:
        return PropertyWpaKeyMgmtValueWpaEapSuiteB;
```

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -119,6 +119,94 @@ HostapdHwModeToPropertyValue(Wpa::HostapdHwMode hwMode)
         throw AccessPointControllerException(std::format("Invalid hostapd hw_mode value {}", magic_enum::enum_name(hwMode)));
     }
 }
+
+std::string
+IeeeAkmSuiteToWpaKeyMgmtPropertyValue(Ieee80211AkmSuite akmSuite)
+{
+    switch (akmSuite) {
+    case Ieee80211AkmSuite::Reserved0:
+        return "";
```
abeltrano, line 128: Should probably remove this and let it go to the `default` case. Same with the others that are returning `""`, with the exception of `Ieee8021xSuiteB`.

Any value not appearing in hostapd's `hostapd_config_parse_key_mgm()` will be rejected. If this function was not hostapd-specific, I'd say leave it, but it'd be better to catch invalid input arguments to hostapd earlier.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -200,6 +288,58 @@ AccessPointControllerLinux::SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol
...
+        isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameWpaKeyMgmt, concatenatedWpaKeyMgmtPropertyValue);
+
+        // Set the hostapd wpa property.
+        isOk = isOk && m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameWpa, Wpa::ProtocolHostapd::PropertyWpaValue2); // TODO: Set this correctly
+    } catch (const Wpa::HostapdException& ex) {
+        throw AccessPointControllerException(std::format("Failed to set AKM suites for interface {} ({})", GetInterfaceName(), ex.what()));
+    }
+
+    // Reload hostapd conf file.
+    return isOk && m_hostapd.Reload();
```
abeltrano, line 314: For ease/clarity during debugging, can we separate the `Reload()` into it's own asssignment?

```suggestion
    // Reload hostapd conf file.
    isOk = isOk && m_hostapd.Reload();

    return isOk;
```

This way you can break on either line and see the current state (`isOk`) and the result of the `Reload` if `isOk` was `true`.

src/linux/wpa-controller/include/Wpa/ProtocolHostapd.hxx
```diff
@@ -200,6 +200,39 @@ struct ProtocolHostapd :
...
+    static constexpr auto PropertyWpaValue2 = "2"; // IEEE 802.11i/RSN (WPA2)
+    static constexpr auto PropertyWpaValue3 = "3"; // WPA and WPA2
+
+    static constexpr auto PropertyNameWpaKeyMgmt = "wpa_key_mgmt";
+    static constexpr auto PropertyWpaKeyMgmtValueWpaPsk = "WPA-PSK";
+    static constexpr auto PropertyWpaKeyMgmtValueWpaPskSha256 = "WPA-PSK-SHA256";
+    static constexpr auto PropertyWpaKeyMgmtValueWpaEap = "WPA-EAP";
+    static constexpr auto PropertyWpaKeyMgmtValueWpaEapSha256 = "WPA-EAP-SHA256";
+    static constexpr auto PropertyWpaKeyMgmtValueSae = "SAE";
+    static constexpr auto PropertyWpaKeyMgmtValueWpaEapSuiteB192 = "WPA-EAP-SUITE-B-192";
```
abeltrano, line 215: Missing one here:

```suggestion
    static constexpr auto PropertyWpaKeyMgmtValueWpaEapSuiteB = "WPA-EAP-SUITE-B";
    static constexpr auto PropertyWpaKeyMgmtValueWpaEapSuiteB192 = "WPA-EAP-SUITE-B-192";
```

tests/unit/wifi/helpers/AccessPointControllerTest.cxx
```diff
@@ -52,6 +52,28 @@ AccessPointControllerTest::SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol i
...
 }
 
+bool
+AccessPointControllerTest::SetAkmSuites(std::vector<Microsoft::Net::Wifi::Ieee80211AkmSuite> akmSuites)
+{
+    if (AccessPoint == nullptr) {
+        throw std::runtime_error("AccessPointControllerTest::SetAkmSuites called with null AccessPoint");
+    }
+
+    // TODO: Set AKM suites.
```
abeltrano, line 62: Should be similar to prior additions where we add a public member to `AccessPointTest` for whatever is being set, so `std::vector<Microsoft::Net::Wifi::Ieee80211AkmSuite>` here, then update that member with the value being set. Similarly for cipher suites.

abeltrano (changes requested):

abeltrano (changes requested):
I've looked at everything except the changes in `WifiCore.proto`. Still trying to sort out whether the encoding matches the spec.

src/common/service/NetRemoteService.cxx
```diff
@@ -187,6 +214,55 @@ Ieee80211AkmSuiteToNetRemoteAkm(Ieee80211AkmSuite akmSuite)
...
+    case Dot11AkmSuite::Dot11AkmSuiteFtFilsSha384:
+        return Ieee80211AkmSuite::FtFilsSha384;
+    case Dot11AkmSuite::Dot11AkmSuiteOwe:
+        return Ieee80211AkmSuite::Owe;
+    case Dot11AkmSuite::Dot11AkmSuiteFtPskSha384:
+        return Ieee80211AkmSuite::FtPskSha384;
+    case Dot11AkmSuite::Dot11AkmSuitePskSha384:
+        return Ieee80211AkmSuite::PskSha384;
+    default:
+        return Ieee80211AkmSuite::Reserved0;
```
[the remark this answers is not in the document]
corbin-phipps: Agreed, will make that change

src/common/service/NetRemoteService.cxx
```diff
@@ -494,6 +602,113 @@ NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerConte
...
+
+    // Convert Dot11AkmSuiteConfigurations to Ieee equivalent values.
+    std::vector<Microsoft::Net::Wifi::Ieee80211AkmSuite> ieeeAkmSuites;
+    std::vector<Microsoft::Net::Wifi::Ieee80211AuthenticationAlgorithm> ieeeAuthenticationAlgorithms;
+    std::ranges::transform(akmSuiteConfigurations, std::back_inserter(ieeeAkmSuites), [&](const Dot11AkmSuiteConfiguration& akmSuiteConfiguration) {
+        return detail::NetRemoteAkmToIeee80211AkmSuite(akmSuiteConfiguration.akmsuite());
+    });
+    std::ranges::transform(akmSuiteConfigurations, std::back_inserter(ieeeAuthenticationAlgorithms), [&](const Dot11AkmSuiteConfiguration& akmSuiteConfiguration) {
+        return detail::NetRemoteAuthenticationAlgorithmToIeeeAuthenticationAlgorithm(akmSuiteConfiguration.authenticationalgorithm());
+    });
```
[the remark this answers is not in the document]
corbin-phipps: Thanks, I've been working on learning more about `std::ranges` and `std::views`, so this is useful knowledge

src/linux/wifi/core/AccessPointControllerLinux.cxx
corbin-phipps: For some reason, I don't see that in the hostapd.conf file that I have, it only has SuiteB192, but perhaps it was just forgotten in the comments for that example conf file

corbin-phipps: Yeah that appears to be the case, since I see both suite B and suite B192 in `config_file.c`

src/linux/wifi/core/AccessPointControllerLinux.cxx
corbin-phipps: Updated for this function, but still need to do something similar for the other helpers. Need to think a bit more about those though

src/common/service/NetRemoteService.cxx
```diff
@@ -494,6 +620,116 @@ NetRemoteService::WifiAccessPointSetPhyType([[maybe_unused]] ::grpc::ServerConte
...
+        auto accessPointCapabilities = accessPointController->GetCapabilities();
+        const auto& supportedAkmSuites = accessPointCapabilities.AkmSuites;
+        const auto& supportedCipherSuites = accessPointCapabilities.CipherSuites;
+
+        if (!std::empty(akmSuiteConfigurations)) {
+            // Remove unsupported AKM suites.
+            std::erase_if(ieeeAkmSuites, [&](const auto& ieeeAkmSuite) {
+                return std::ranges::find(supportedAkmSuites, ieeeAkmSuite) == std::cend(supportedAkmSuites);
+            });
+            // TODO: Do the same for other parts of Dot11AkmSuiteConfiguration.
```
abeltrano, line 694: Are there actually any parts left to do?

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -197,7 +285,72 @@ AccessPointControllerLinux::SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol
...
+{
+    bool isOk = false;
+
+    // Construct the wpa_key_mgmt hostapd property value.
+    std::string concatenatedWpaKeyMgmtPropertyValue{};
+    for (const auto& akmSuite : akmSuites) {
+        const auto& wpaKeyMgmtPropertyValue = detail::IeeeAkmSuiteToWpaKeyMgmtPropertyValue(akmSuite);
+        concatenatedWpaKeyMgmtPropertyValue += wpaKeyMgmtPropertyValue;
+        concatenatedWpaKeyMgmtPropertyValue += " ";
+    }
```
abeltrano, line 304: `std::string+=` isn't great for concatenation. For a one-liner it's ok, though even then `std::format()` is probably a bit better. For repeated concatenations like this, `std::ostringstream` will be a lot better.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -197,7 +285,72 @@ AccessPointControllerLinux::SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol
...
+
+    try {
+        // Set the hostapd wpa_key_mgmt property.
+        isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameWpaKeyMgmt, concatenatedWpaKeyMgmtPropertyValue);
+
+        // For simplicity, set the hostapd wpa property to allow WPA and WPA2 (and WPA3).
+        isOk = isOk && m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameWpa, Wpa::ProtocolHostapd::PropertyWpaValueWpaAndWpa2);
+
+        // For simplicity, set the hostapd auth_algs property to allow both Open System Authentication and Shared Key Authentication.
+        isOk = isOk && m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameAuthAlgs, Wpa::ProtocolHostapd::PropertyAuthAlgsValueOpenAndSharedKey);
```
abeltrano, line 314: Is the intention to do something different in the future, or, just leave this as-is? If the former, let's add a comment.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -197,7 +285,72 @@ AccessPointControllerLinux::SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol
...
+
+    return isOk;
+}
+
+bool
+AccessPointControllerLinux::SetCipherSuites(std::vector<Ieee80211CipherSuite> cipherSuites)
+{
+    bool isOk = false;
+
+    // TODO: Handle WEP separately.
```
abeltrano, line 330: Let's just ignore WEP for now.

src/linux/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -197,7 +285,72 @@ AccessPointControllerLinux::SetProtocol(Microsoft::Net::Wifi::Ieee80211Protocol
...
+        concatenatedPairwisePropertyValue += pairwisePropertyValue;
+        concatenatedPairwisePropertyValue += " ";
+    }
+
+    try {
+        // Set the hostapd wpa_pairwise property.
+        isOk = m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertyNameWpaPairwise, concatenatedPairwisePropertyValue);
+
+        // Set the hostapd rsn_pairwise property.
+        isOk = isOk && m_hostapd.SetProperty(Wpa::ProtocolHostapd::PropertynameRsnPairwise, concatenatedPairwisePropertyValue);
```
abeltrano, line 345: Just want to confirm that both `WpaPairWise` and `RsnPairWise` are supposed to get the same value here? Also, it looks like `PropertynameRsnPairwise` should be renamed to `PropertyNameRsnPairwise` (lowercase 'n' -> 'N').

src/common/service/NetRemoteService.cxx
corbin-phipps: Yes, we still need the configuration values (shared key, radius server configuration, or empty) that come from `oneof Configuration` in `WifiCore.proto`, which also need to be tightly coupled with the AKM suite. I'm working on some changes to do this, but I'm struggling a bit...

src/linux/wifi/core/AccessPointControllerLinux.cxx
corbin-phipps: I was thinking of leaving it as-is, but I suppose there could be a situation where, for example, a client only wants to use the original WPA. Currently, with this set the way it's set now, hostapd will likely choose the best WPA version that it can since all are enabled

src/linux/wifi/core/AccessPointControllerLinux.cxx
corbin-phipps: The default value for `rsn_pairwise` in hostapd is the value of `wpa_pairwise`, so I just set them both to the same value for simplicity. However, they could technically be set with different values, so perhaps this should be changed. And good catch!

PR #174, opened 2024-02-26
Add data streaming service with WifiDataStreamUpload API

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR adds a new gRPC service for data streaming to NetRemote, along with one API method for client-to-server streaming.

### Technical Details

* Add `NetRemoteDataStreamingService` gRPC service and run from `NetRemoteServer`.
* Add `WifiDataStreamUpload` API to `NetRemoteDataStreamingService.proto`.
* Implement `WifiDataStreamUpload` API in `NetRemoteDataStreamingService`, factoring out the reactor class into `NetRemoteDataStreamingReactors`.
* Add unit tests for new API.

### Test Results

All tests pass.

### Reviewer Focus

None.

### Future Work

None.

### Checklist


protocol/protos/NetRemoteDataStreamingService.proto
```diff
@@ -0,0 +1,12 @@
...
+syntax = "proto3";
+
+package Microsoft.Net.Remote.Service;
+
+import "NetRemote.proto";
+import "NetRemoteWifi.proto";
+
+service NetRemoteDataStreaming
+{
+    rpc WifiDataStreamUpload (stream Microsoft.Net.Remote.Wifi.WifiDataStreamUploadData) returns (Microsoft.Net.Remote.Wifi.WifiDataStreamUploadResult);
```
abeltrano, line 11: Is there any reason this needs to be wifi specific? Calling this on a connection using a non-wifi adapter would still work fine, right? The same goes to all the proposed new types that reference `Wifi`. I think we can make these core streaming types independent of wifi.

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
```diff
@@ -0,0 +1,136 @@
...
+        std::vector<std::unique_ptr<DataStreamWriter>> dataStreamWriters;
+        std::vector<std::jthread> clientThreads;
+
+        for (std::size_t i = 0; i < numberOfClients; i++) {
+            dataStreamingClients.push_back(NetRemoteDataStreaming::NewStub(channel));
+            dataStreamWriters.push_back(std::make_unique<DataStreamWriter>(dataStreamingClients.back().get(), numberOfDataBlocksToWrite));
+
+            clientThreads.emplace_back([&, i]() {
+                WifiDataStreamUploadResult result{};
+                grpc::Status status = dataStreamWriters[i]->Await(&result);
```
abeltrano, line 129: You can introduce a new variable in the capture here to make accessing the stream writer a bit simpler. Also, it's safer to move the data accessed by the thread into the thread func itself if it's not shared (which it appears not to be IIUC), effectively transferring ownership of it and then you don't need to worry about its lifetime. At that point, it doesn't seem like the `dataStreamWriters` `vector` is needed at all.

```suggestion
            auto dataStreamWriter = std::make_unique<DataStreamWriter>(dataStreamingClients.back().get(), numberOfDataBlocksToWrite);
            clientThreads.emplace_back([dataStreamWriter = std::move(dataStreamWriter)]() {
                WifiDataStreamUploadResult result{};
                grpc::Status status = dataStreamWriter->Await(&result);
```

You could take that even further and only pass a ref to the `channel` into the capture and create both the client and writer inside the thread func.

abeltrano (changes requested):

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
```diff
@@ -0,0 +1,136 @@
...
+
+        grpc::Status
+        Await(WifiDataStreamUploadResult* result)
+        {
+            std::unique_lock lock(m_mutex);
+
+            m_cv.wait(lock, [this] {
+                return m_done;
+            });
+            *result = m_result;
```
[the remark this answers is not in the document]
corbin-phipps: The caller of `Await` is the `DataStreamWriter` that is associated with a specific client, and `m_result` is filled when that client invokes the `WifiDataStreamUpload` API, so it should be fine to not clear `m_result`.

I will move the helper class to a separate file with comments

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
corbin-phipps: Good idea, I was having lifetime management issues which led to me storing the clients and the stream writers into vectors, but I like the idea of moving the data into the lambda instead. I'll try it out

src/common/service/NetRemoteDataStreamingService.cxx
```diff
@@ -0,0 +1,12 @@
+
+#include "NetRemoteDataStreamingReactors.hxx"
+#include <microsoft/net/remote/NetRemoteDataStreamingService.hxx>
+
+using namespace Microsoft::Net::Remote::Service;
+using namespace Microsoft::Net::Remote::Wifi;
+
+grpc::ServerReadReactor<WifiDataStreamUploadData>*
+NetRemoteDataStreamingService::WifiDataStreamUpload([[maybe_unused]] grpc::CallbackServerContext* context, WifiDataStreamUploadResult* result)
+{
```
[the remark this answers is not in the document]
corbin-phipps: Will create an issue to track this and will add soon.

src/common/service/NetRemoteDataStreamingService.cxx
```diff
@@ -0,0 +1,12 @@
...
+    return new Reactors::DataStreamReader(result);
```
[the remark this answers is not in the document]
corbin-phipps: Will create an issue to track this and will add soon.

protocol/protos/NetRemoteDataStream.proto
```diff
@@ -0,0 +1,31 @@
+
+syntax = "proto3";
+
+package Microsoft.Net.Remote.DataStream;
+
+import "google/protobuf/any.proto";
```
abeltrano, line 6: This isn't needed now, right?

tests/unit/TestNetRemoteDataStreamingReactors.cxx
```diff
@@ -0,0 +1,60 @@
...
+
+grpc::Status
+DataStreamWriter::Await(DataStreamUploadResult* result)
+{
+    std::unique_lock lock(m_writeStatusGate);
+    static constexpr auto timeoutValue = std::chrono::seconds(10);
+
+    m_writesDone.wait_for(lock, timeoutValue, [this] {
+        return m_done;
+    });
```
abeltrano, line 44: Need to check this result for timeout status:

```suggestion
    const auto waitResult = m_writesDone.wait_for(lock, timeoutValue, [this] {
        return m_done;
    });
    if (waitResult == std::cv_status::timeout) {
        // handle timeout
```

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
```diff
@@ -0,0 +1,68 @@
...
+
+            clientThreads.emplace_back([dataStreamWriter = std::move(dataStreamWriter)]() {
+                DataStreamUploadResult result{};
+                grpc::Status status = dataStreamWriter->Await(&result);
+                REQUIRE(status.ok());
+                REQUIRE(result.numberofdatablocksreceived() == numberOfDataBlocksToWrite);
+                REQUIRE(result.status().code() == DataStreamOperationStatusCodeSucceeded);
+            });
+        }
+    }
```
abeltrano, line 67: I think we should verify that the streaming threads don't throw any exceptions. Right now, when the threads are destroyed at end of scope, their destructors will call `join()` automatically, but if an exception is thrown, `std::terminate` will be called and the test will crash. Instead, we should call `join` explicitly and verify no exceptions are thrown.

```suggestion
        }
        for (auto& clientThread : clientThreads) {
            REQUIRE_NOTHROW(clientThread.join());
        }
    }
```

abeltrano (changes requested):

protocol/protos/NetRemoteDataStreamingService.proto
```diff
@@ -0,0 +1,12 @@
...
+syntax = "proto3";
+
+package Microsoft.Net.Remote.Service;
+
+import "NetRemote.proto";
+import "NetRemoteDataStream.proto";
+
+service NetRemoteDataStreaming
+{
+    rpc DataStreamUpload (stream Microsoft.Net.Remote.DataStream.DataStreamUploadData) returns (Microsoft.Net.Remote.DataStream.DataStreamUploadResult);
```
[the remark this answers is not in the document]
corbin-phipps: I wasn't planning to for this Upload API since the client is in full control - the server simply reads data until the client stops writing, which is determined by the client. I don't think the client needs to specify a "fixed vs continuous" streaming type to the server since the client will just send what it wants.

For the Download API however, the client would request a streaming type and the server would send data according to that requested type - the client this time would read data until the server stops writing, which again is determined by the client (via the fixed number of blocks requested by the client or via an RPC cancellation by the client in the continuous case).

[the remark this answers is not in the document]
abeltrano: Good point, I guess the sever is a pure sink in this case. Thanks!

src/common/service/NetRemoteDataStreamingService.cxx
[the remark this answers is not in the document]
corbin-phipps: #179

src/common/service/NetRemoteDataStreamingService.cxx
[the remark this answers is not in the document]
corbin-phipps: #180

tests/unit/TestNetRemoteDataStreamingReactors.cxx
```diff
@@ -0,0 +1,67 @@
...
+    std::unique_lock lock(m_writeStatusGate);
+    static constexpr auto timeoutValue = 10s;
+
+    const auto isDone = m_writesDone.wait_for(lock, timeoutValue, [this] {
+        return m_done;
+    });
+
+    if (!isDone) {
+        DataStreamOperationStatus status{};
+        status.set_code(DataStreamOperationStatusCode::DataStreamOperationStatusCodeFailed);
```
abeltrano, line 48: Would a ne wstatus code like `DataStreamOperationStatusCode::DataStreamOperationStatusTimedOut` make sense here?

corbin-phipps: I think that's a good idea. Probably best to have the status codes be as specific as possible and leave the "failed" code for failures without a clear cause

tests/unit/TestNetRemoteDataStreamingReactors.hxx
```diff
@@ -0,0 +1,75 @@
...
+#define TEST_NET_REMOTE_DATA_STREAMING_REACTORS_HXX
+
+#include <condition_variable>
+#include <mutex>
+
+#include <microsoft/net/remote/protocol/NetRemoteDataStreamingService.grpc.pb.h>
+
+namespace Microsoft::Net::Remote::Test
+{
+using namespace std::chrono_literals;
```
abeltrano, line 12: This should move to the source file (with `#include <chrono>`) since it's not used here.

tests/unit/TestNetRemoteDataStreamingReactors.hxx
```diff
@@ -0,0 +1,75 @@
+
+#ifndef TEST_NET_REMOTE_DATA_STREAMING_REACTORS_HXX
+#define TEST_NET_REMOTE_DATA_STREAMING_REACTORS_HXX
+
+#include <condition_variable>
```
abeltrano, line 5: Missing header for `uint32_t`:
```suggestion
#include <condition_variable>
#include <cstdint>
```

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
```diff
@@ -0,0 +1,72 @@
+
+#include <condition_variable>
+#include <memory>
+#include <mutex>
+#include <thread>
```
abeltrano, line 5: Missing header for `std::move`:

```suggestion
#include <thread>
#include <utility>
```

tests/unit/TestNetRemoteDataStreamingReactors.hxx
```diff
@@ -0,0 +1,75 @@
...
+#include <mutex>
+
+#include <microsoft/net/remote/protocol/NetRemoteDataStreamingService.grpc.pb.h>
```
abeltrano, line 8: Also need `NetRemoteDataStream.pb.h` for some types here:

```suggestion
#include <microsoft/net/remote/protocol/NetRemoteDataStream.pb.h>
#include <microsoft/net/remote/protocol/NetRemoteDataStreamingService.grpc.pb.h>
```

tests/unit/TestNetRemoteDataStreamingReactors.cxx
```diff
@@ -0,0 +1,67 @@
+
```
abeltrano, line 1: Missing header for `std::format`:
```suggestion
#include <format>

```

abeltrano (approved):

Issue #179, opened 2024-02-27
Avoid manual memory management in data streaming service


corbin-phipps:
**Is your feature request related to a problem? Please describe.**
The `NetRemoteDataStreamingService` currently uses `new` and `delete` on the `ServerReadReactor`, which may lead to memory management issues.

**Describe the solution you'd like**
Allocate memory with `std::shared_ptr`, such as `std::shared_ptr<grpc::ServerReadReactor<DataStreamUploadData>>`

PR #184, opened 2024-02-28
Use smart pointers for memory management

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [X] Non-functional change

### Goals

Addresses #179 . Manual memory management with `new` and `delete` should be avoided. Memory should be managed with smart pointers instead.

### Technical Details

* Replace `new` in `DataStreamUpload` implementation and instead use `std::unique_ptr`.
* Remove `delete this` from `DataStreamReader` reactor class `OnDone` callback.

### Test Results

All tests pass.

### Reviewer Focus

None.

### Future Work

This currently uses a `std::vector` to store a new reactor for each invocation of `DataStreamUpload` in order to separate clients. However, this would ideally be stored in a `std::unordered_map` instead, with a unique key for each client. Unfortunately, there is no clear way for the service to uniquely identify a client (`context->peer()` returns the same URI for all clients in the unit test).

### Checklist

### DIFF
--- a/src/common/service/NetRemoteDataStreamingService.cxx
+++ b/src/common/service/NetRemoteDataStreamingService.cxx
@@ -12,5 +12,7 @@ NetRemoteDataStreamingService::DataStreamUpload([[maybe_unused]] grpc::CallbackS
 {
     const NetRemoteApiTrace traceMe{};
 
-    return new Reactors::DataStreamReader(result);
+    m_dataStreamReaders.push_back(std::make_unique<Reactors::DataStreamReader>(result));
+
+    return m_dataStreamReaders.back().get();
 }
--- a/src/common/service/NetRemoteDataStreamingReactors.cxx
+++ b/src/common/service/NetRemoteDataStreamingReactors.cxx
@@ -40,5 +40,4 @@ DataStreamReader::OnCancel()
 void
 DataStreamReader::OnDone()
 {
-    delete this;
 }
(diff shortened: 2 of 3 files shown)


src/common/service/NetRemoteDataStreamingService.cxx
```diff
@@ -12,5 +12,7 @@ NetRemoteDataStreamingService::DataStreamUpload([[maybe_unused]] grpc::CallbackS
 {
     const NetRemoteApiTrace traceMe{};
 
-    return new Reactors::DataStreamReader(result);
+    m_dataStreamReaders.push_back(std::make_unique<Reactors::DataStreamReader>(result));
```
abeltrano, line 15: It looks like this memory is only freed when the `NetRemoteDataStreamingService` instance is destroyed, and that means these will accumulate for the lifetime of the server.

I expected the service would be notified in some way that the RPC was complete and could then release its own reference on the reactor but reading the (embarrassingly sparse) documentation and examples, it seems this isn't the case. I'm surprised that gRPC exposes such a brittle design.

In the end, if we don't see a reason `NetRemoteDataStreamingService` needs to account for in-flight RPCs, then your original code (with `delete this` in the `OnDone()` function), should probably be used instead.

corbin-phipps: The service essentially completes the RPC when it calls `Finish` inside the reactor class, which leads to the `OnDone` callback being invoked to allow for any cleanup. This was why I had `delete this` in that callback

abeltrano: I could see why `delete this` was used, but that kind of usage makes me twitchy because we have to trust the calling code is going to respect that the memory is freed after `OnDone`. Hence I thought we could make it safer by using ref-counting, but I hadn't looked at the gRPC API holistically, and erroneously assumed there was some communication to the service about RPC completion status.

These kinds of implicit contracts, especially when undocumented, could change and/or be abused at any time and we'd have no real way of preventing it from happening, possibly causing very bad stuff to happen (ie. silent memory corruption due to accessing freed memory).

PR #185, opened 2024-02-28
Add initial DataStreamDownload API implementation

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR adds the initial implementation of the `DataStreamDownload` API, including the API itself, the foundational service implementation and reactor class, a test client reactor class, and a basic unit test.

### Technical Details

* Added `DataStreamDownload` API to `NetRemoteDataStreamingService.proto`, along with applicable types into `NetRemoteDataStream.proto`.
* Added `DataStreamWriter` implementation of the gRPC `ServerWriteReactor` reactor class to `NetRemoteDataStreamingReactors`.
* Added test client implementation of the gRPC `ClientReadReactor` reactor class.
* Added basic unit test.
* Reverted smart pointer memory management back to `new`/`delete`.

### Test Results

All tests pass.

### Reviewer Focus

None.

### Future Work

* Handle `DataStreamPattern` options from the client's `DataStreamDownloadRequest`.
* Add additional unit test for `DataStreamTypeContinuous` with `DataStreamPatternConstant`.
* Add a better mechanism for data creation.

### Checklist

### DIFF
--- a/protocol/protos/NetRemoteDataStream.proto
+++ b/protocol/protos/NetRemoteDataStream.proto
@@ -23,8 +23,52 @@ message DataStreamUploadData
     bytes Data = 1;
 }
 
+message DataStreamDownloadData
+{
+    bytes Data = 1;
+    uint32 SequenceNumber = 2;
+    DataStreamOperationStatus Status = 3;
+}
+
 message DataStreamUploadResult
 {
     uint32 NumberOfDataBlocksReceived = 1;
     DataStreamOperationStatus Status = 2;
+}
+
+enum DataStreamType
+{
+    Fixed = 0;
+    Continuous = 1;
+}
+
+enum DataStreamPattern
+{
+    Constant = 0;
+}
+
+message DataStreamFixedTypeProperties
+{
+    uint32 NumberOfDataBlocksToStream = 1;
+}
+
+message DataStreamContinuousTypeProperties
+{
+
+}
+
+message DataStreamProperties
+{
+    DataStreamType Type = 1;
+    DataStreamPattern Pattern = 2;
+    oneof Properties
+    {
+        DataStreamFixedTypeProperties FixedTypeProperties = 3;
+        DataStreamContinuousTypeProperties ContinuousTypeProperties = 4;
+    }
+}
+
+message DataStreamDownloadRequest
+{
+    DataStreamProperties Properties = 1;
 }
--- a/protocol/protos/NetRemoteDataStream.proto
+++ b/protocol/protos/NetRemoteDataStream.proto
@@ -38,13 +38,15 @@ message DataStreamUploadResult
 
 enum DataStreamType
 {
-    Fixed = 0;
-    Continuous = 1;
+    DataStreamTypeUnknown = 0;
+    DataStreamTypeFixed = 1;
+    DataStreamTypeContinuous = 2;
 }
 
 enum DataStreamPattern
 {
-    Constant = 0;
+    DataStreamPatternUnknown = 0;
+    DataStreamPatternConstant = 1;
 }
 
 message DataStreamFixedTypeProperties
--- a/src/common/service/NetRemoteDataStreamingReactors.cxx
+++ b/src/common/service/NetRemoteDataStreamingReactors.cxx
@@ -102,6 +102,8 @@ DataStreamWriter::OnWriteDone(bool isOk)
 void
 DataStreamWriter::OnCancel()
 {
+    // The RPC is cancelled, so call Finish to complete it.
+    Finish(grpc::Status::CANCELLED);
 }
 
 void
--- a/src/common/service/NetRemoteDataStreamingReactors.cxx
+++ b/src/common/service/NetRemoteDataStreamingReactors.cxx
@@ -107,7 +107,7 @@ DataStreamWriter::OnWriteDone(bool isOk)
 void
 DataStreamWriter::OnCancel()
 {
-    // The RPC is cancelled, so call Finish to complete it.
+    // The RPC is cancelled by the client, so call Finish to complete it from the server perspective.
     Finish(grpc::Status::CANCELLED);
 }
 
--- a/src/common/service/NetRemoteDataStreamingReactors.cxx
+++ b/src/common/service/NetRemoteDataStreamingReactors.cxx
@@ -43,6 +43,7 @@ DataStreamReader::OnCancel()
 void
 DataStreamReader::OnDone()
 {
+    delete this;
 }
 
 DataStreamWriter::DataStreamWriter(const DataStreamDownloadRequest* request)
@@ -109,6 +110,7 @@ DataStreamWriter::OnCancel()
 void
 DataStreamWriter::OnDone()
 {
+    delete this;
 }
 
 void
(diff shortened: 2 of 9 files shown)


protocol/protos/NetRemoteDataStream.proto
```diff
@@ -23,8 +23,54 @@ message DataStreamUploadData
...
+
+message DataStreamProperties
+{
+    DataStreamType Type = 1;
+    DataStreamPattern Pattern = 2;
+    oneof Properties
+    {
+        DataStreamFixedTypeProperties FixedTypeProperties = 3;
+        DataStreamContinuousTypeProperties ContinuousTypeProperties = 4;
+    }
```
abeltrano, line 70: We should de-dupe `Properties` from the `oneof` field names. Also, `Type` probably isn't needed either:

```proto
    oneof Properties
    {
        DataStreamFixedTypeProperties Fixed = 3;
        DataStreamContinuousTypeProperties Continuous = 4;
    }
```

src/common/service/NetRemoteDataStreamingReactors.cxx
```diff
@@ -40,4 +43,107 @@ DataStreamReader::OnCancel()
...
+        *m_data.mutable_status() = m_writeStatus;
+        StartWrite(&m_data);
+    } else {
+        // No more data to write.
+        Finish(::grpc::Status::OK);
+    }
+}
+
+void
+DataStreamWriter::HandleFailure(const std::string errorMessage)
```
abeltrano, line 140: This is accepting a copy of the string, and then I believe another copy is made when passed to `set_message`. It probably makes sense to accept a `const std::string&` here instead, which will avoid 1 of those copies.

Also, in general, when accepting a copy/value of an argument, it's usually a good idea to accept it as non-`const` so the function can use it however it likes, since it owns that memory.

tests/unit/TestNetRemoteDataStreamingReactors.cxx
```diff
@@ -70,3 +70,62 @@ DataStreamWriter::NextWrite()
...
+    m_status = status;
+    m_done = true;
+    m_readsDone.notify_one();
+}
+
+grpc::Status
+DataStreamReader::Await(uint32_t* numberOfDataBlocksReceived, DataStreamOperationStatus* operationStatus)
+{
+    std::unique_lock lock(m_readStatusGate);
+    static constexpr auto timeoutValue = 10s;
```
abeltrano, line 105: I think we'll want to eventually use these test classes outside test code (and move them to a common lib at that time).

In that case, it's likely we'll want to make this timeout value configurable. Suggest to at least add a member to `DataStreamMember` and read the member here instead of using the constant. Completely fine if the member is just initialized to a constant (with this value) and no way is provided for configuring it for now.

protocol/protos/NetRemoteDataStream.proto
```diff
@@ -23,8 +23,54 @@ message DataStreamUploadData
...
 }
+
+enum DataStreamType
+{
+    DataStreamTypeUnknown = 0;
+    DataStreamTypeFixed = 1;
+    DataStreamTypeContinuous = 2;
+}
+
+enum DataStreamPattern
```
abeltrano, line 46: Does this refer to the throughput pattern, for example, the timing/rate of the data flow? In that case, `Constant` == an uninterrupted flow of data, correct?

protocol/protos/NetRemoteDataStream.proto
```diff
@@ -23,8 +23,54 @@ message DataStreamUploadData
     bytes Data = 1;
 }
 
+message DataStreamDownloadData
+{
+    bytes Data = 1;
+    uint32 SequenceNumber = 2;
+    DataStreamOperationStatus Status = 3;
```
abeltrano, line 30: Is `Data` only present/valid when `Status == Succeeded`? If so, suggest re-ordering these such that always-present/valid fields are first, then optional/conditional ones following.

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
```diff
@@ -71,3 +72,48 @@ TEST_CASE("DataStreamUpload API", "[basic][rpc][client][remote][stream]")
...
+
+        DataStreamProperties properties{};
+        properties.set_type(DataStreamType::DataStreamTypeFixed);
+        properties.set_pattern(DataStreamPattern::DataStreamPatternConstant);
+        *properties.mutable_fixedtypeproperties() = std::move(fixedTypeProperties);
+
+        DataStreamDownloadRequest request{};
+        *request.mutable_properties() = std::move(properties);
+
+        auto dataStreamReader = std::make_unique<DataStreamReader>(client.get(), &request);
```
abeltrano, line 110: Any reason this can't be allocated on the stack instead? Eg.

```C++
DataStreamReader dataStreamReader{  client.get(), &request };
// ...
grpc::Status status = dataStreamReader.Await(&numberOfDataBlocksReceived, &operationStatus);
```
```

File: tests/unit/TestNetRemoteDataStreamingReactors.cxx [hunk 7]
```diff
@@ -70,3 +70,62 @@ DataStreamWriter::NextWrite()
...
+        DataStreamOperationStatus status{};
+        status.set_code(DataStreamOperationStatusCode::DataStreamOperationStatusCodeTimedOut);
+        status.set_message("Timeout occurred while waiting for all reads to be completed");
+        *m_data.mutable_status() = std::move(status);
+    }
+
+    // Handle mismatched sequence number and number of data blocks received.
+    if (m_data.sequencenumber() != m_numberOfDataBlocksReceived) {
+        DataStreamOperationStatus status{};
+        status.set_code(DataStreamOperationStatusCode::DataStreamOperationStatusCodeFailed);
```
abeltrano, line 122: Looking at this now, would it be helpful to know when mismatched sequencing occurs? Might this allow us to help quantify the number of blocks dropped/lost?

protocol/protos/NetRemoteDataStream.proto
corbin-phipps: Right, `Constant` would be a steady pattern with each data message being the same number of bytes

protocol/protos/NetRemoteDataStream.proto
corbin-phipps: Yes that's true. I can re-order the fields so that `Status` is first since that will always be present.

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
corbin-phipps: I believe it can be stack allocated. Will update

tests/unit/TestNetRemoteDataStreamingReactors.cxx
corbin-phipps: That's a good idea. It would be easy to tell - basically move this check into `OnReadDone`. We'd need to decide if we want to stop reading and cancel the RPC as soon as the first sequencing mismatch occurs, or if we want to keep going and later count how many blocks were dropped.

corbin-phipps: Going to defer this until a bit later since it's test code - tracked by #186

Issue #186, opened 2024-02-29
Check when mismatched sequencing occurs in DataStreamDownload test client reactor


corbin-phipps:
**Is your feature request related to a problem? Please describe.**
In `TestNetRemoteDataStreamingReactors`, the `Await` function currently checks if the sequence number matches the number of data blocks received. However, it would be helpful to know when exactly that sequence mismatch occurs.

**Describe the solution you'd like**
Check for sequence mismatch in `OnReadDone` callback and handle accordingly. Consider either stopping the RPC at the first mismatch, or continuing reads and tracking the number of data blocks dropped/lost.

PR #202, opened 2024-03-02
Add DataStreamBidirectional API

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR adds the `DataStreamBidirectional` API along with a unit test.

### Technical Details

* Added `DataStreamBidirectional` API to `NetRemoteDataStreamingService.proto`.
* Added `DataStreamReaderWriter` implementation of the gRPC `ServerBidiReactor` to `NetRemoteDataStreamingReactors`.
* Added `DataStreamReaderWriter` implementation of the gRPC `ClientBidiReactor` to `TestNetRemoteDataStreamingReactors`.
* Added unit test to `TestNetRemoteDataStreamingServiceClient`.

### Test Results

All tests pass.

### Reviewer Focus

None.

### Future Work

* Apply the work for #186 to this API as well.
* Check if access/writes to `m_status` need to be gated.
* Consider if the client will ever need to cancel this RPC and if so, add `Cancel()` function and another unit test.

### Checklist

### DIFF
--- a/src/common/service/NetRemoteDataStreamingReactors.cxx
+++ b/src/common/service/NetRemoteDataStreamingReactors.cxx
@@ -235,3 +235,45 @@ DataStreamWriter::HandleFailure(const std::string& errorMessage)
     // DataStreamOperationStatusCodeFailed status code set here to know to complete the RPC.
     StartWrite(&m_data);
 }
+
+DataStreamReaderWriter::DataStreamReaderWriter()
+{
+    const FunctionTracer traceMe{};
+}
+
+void
+DataStreamReaderWriter::OnReadDone(bool isOk)
+{
+    const FunctionTracer traceMe{};
+}
+
+void
+DataStreamReaderWriter::OnWriteDone(bool isOk)
+{
+    const FunctionTracer traceMe{};
+}
+
+void
+DataStreamReaderWriter::OnCancel()
+{
+    const FunctionTracer traceMe{};
+}
+
+void
+DataStreamReaderWriter::OnDone()
+{
+    const FunctionTracer traceMe{};
+    delete this;
+}
+
+void
+DataStreamReaderWriter::NextWrite()
+{
+    const FunctionTracer traceMe{};
+}
+
+void
+DataStreamReaderWriter::HandleFailure(const std::string& errorMessage)
+{
+    const FunctionTracer traceMe{};
+}
--- a/src/common/service/NetRemoteDataStreamingReactors.cxx
+++ b/src/common/service/NetRemoteDataStreamingReactors.cxx
@@ -259,7 +259,10 @@ DataStreamReaderWriter::OnReadDone(bool isOk)
     } else {
         // A false "isOk" value could either mean a failed RPC or that no more data is available.
         // Unfortunately, there is no clear way to tell which situation occurred.
-        Finish(grpc::Status::OK);
+        bool readsDoneExpected{ false };
+        if (m_readsDone.compare_exchange_strong(readsDoneExpected, true, std::memory_order_relaxed, std::memory_order_relaxed)) {
+            Finish(grpc::Status::OK);
+        }
     }
 }
 
@@ -287,7 +290,11 @@ DataStreamReaderWriter::OnWriteDone(bool isOk)
         m_status.set_message("Data write successful");
         NextWrite();
     } else {
-        //HandleFailure("Data write failed");
+        // If OnReadDone() failed (due to no more data sent from the client), then Finish() was called and this write will fail.
+        // In that case, do not call HandleFailure() because that triggers another write. Otherwise, this is a true write failure.
+        if (!m_readsDone.load(std::memory_order_relaxed)) {
+            HandleFailure("Data write failed");
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/common/service/NetRemoteDataStreamingReactors.cxx
```diff
@@ -235,3 +235,121 @@ DataStreamWriter::HandleFailure(const std::string& errorMessage)
...
+
+    // Client may have canceled the RPC, so check for cancelation to prevent writing more data
+    // when we shouldn't.
+    if (m_isCanceled.load(std::memory_order_relaxed)) {
+        LOGD << "RPC canceled, aborting write";
+        return;
+    }
+
+    // Create data to write to the client.
+    const auto data = m_dataGenerator.GenerateRandomData();
```
abeltrano, line 333: Are there plans to allow configuration of the block size?

tests/unit/TestNetRemoteDataStreamingReactors.hxx
```diff
@@ -136,6 +136,76 @@ private:
     std::chrono::duration<uint32_t> m_readsDoneTimeoutValue{ DefaultTimeoutValue };
 };
 
+/**
+ * @brief Implementation of the gRPC ClientBidiReactor for client-side data stream reading and writing.
+ */
+class DataStreamReaderWriter : public grpc::ClientBidiReactor<Microsoft::Net::Remote::DataStream::DataStreamUploadData, Microsoft::Net::Remote::DataStream::DataStreamDownloadData>
```
abeltrano, line 142: nit: style (I think clang-format missed this for some reason)
```suggestion
class DataStreamReaderWriter : 
    public grpc::ClientBidiReactor<Microsoft::Net::Remote::DataStream::DataStreamUploadData, Microsoft::Net::Remote::DataStream::DataStreamDownloadData>
```

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
```diff
@@ -1,4 +1,5 @@
 
+#include <bits/chrono.h>
```
abeltrano, line 2: I know clang-tidy complains without this, however, I'm fairly certain it's a false positive and this header doesn't exist on Windows. Fixing it on a per-platform, per-compiler basis is way more trouble than it's worth, so I'm happy to just live with the warning, or, disable that warning for this file (`NOLINTBEGIN(misc-include-cleaner) ... NOLINTEND(misc-include-cleaner)`). Either way, let's add the actual header needed here:

```suggestion
#include <chrono>
```

abeltrano (approved):

src/common/service/NetRemoteDataStreamingReactors.cxx
corbin-phipps: Not currently. Because this API doesn't have a "request"/"response" like regular RPCs and instead has a "stream" for both directions, there is no intuitive way to allow the client to specify a block size that it wants the server to use. I think that if there was a scenario that required that, then the DataStreamDownload API should be used instead

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
corbin-phipps: I initially added `<chrono>` but then clang-tidy complained about that 😆 it says that the `<chrono>` header is not used directly

abeltrano: Yeah, it's a false positive. Fine to ignore or disable in this case.

src/common/service/NetRemoteDataStreamingReactors.cxx
abeltrano: If this is a requirement (I think it is), there are a few ways we could consider achieving this:

1. Since all fields are optional in proto3 format, the first input message from the client could populate a field that specifies properties about the requested stream from the server. This field is only inspected on the first message received and otherwise not even checked for.
2. Streaming could be stateful, where a request is first made to create a stream, and then that stream has to be explicitly started with the usual API. This way, both the client and server can negotiate stream properties. It could look something like this:

```proto
enum DataStreamDirection
{
    Upload,
    Download,
    Both,
};
enum DataStreamCreateStatusCode 
{
    Unknown = 0,
    Succeeded = 1,
    Failed = 2,
    // Others as needed
};
message DataStreamCreateRequest
{
    DataStreamDirection Direction = 1;
    DataStreamProperties PropertiesUpload = 2; // present when Direction == Upload and Both
    DataStreamProperties PropertiesDownload = 2; // present when Direction == Download and Both
};
message DataStreamCreateResult
{
    DataStreamCreateStatusCode Status = 0;
    string StreamId = 1; // Valid/present only when  Status == Succeeded
};
message DataStreamDownloadRequest
{
    string StreamId = 0;
};
rpc DataStreamCreate (DataStreamCreateRequest request, DataStreamCreateResult result);
```

corbin-phipps: Tracking this in #206 and will look into it later. Your first suggestion is definitely simpler and stays closer to the original design, but your second suggestion is cleaner and simple to understand, although requires a bit more work and changes the design a bit.

abeltrano: Thanks! Both require logic on the server side to check for the presence of a field, and while intuitive for the client, having conditional logic like that feels wrong. Anyway, we can drill into this as part of #206 after we determine if this is really a requirement of not.

PR #205, opened 2024-03-04
Check when mismatched sequencing occurs

### LINKED ISSUE #186: Check when mismatched sequencing occurs in DataStreamDownload test client reactor

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

Fixes #186 

### Technical Details

* Check for mismatched sequencing in `OnReadDone` callbacks and store a list of the unreceived sequence numbers.
* Return a `std::span` parameter of the sequence numbers in the client test `Await` functions.
* Don't set status to failed if sequence mismatch occurs, just let client handle it how it wants to.

### Test Results

All tests pass.

### Reviewer Focus

None.

### Future Work

None.

### Checklist

### DIFF
--- a/tests/unit/TestNetRemoteDataStreamingReactors.cxx
+++ b/tests/unit/TestNetRemoteDataStreamingReactors.cxx
@@ -1,7 +1,9 @@
 
+#include <cstddef>
 #include <cstdint>
 #include <format>
 #include <mutex>
+#include <span>
 #include <utility>
 
 #include <grpcpp/impl/codegen/status.h>
--- a/tests/unit/TestNetRemoteDataStreamingReactors.cxx
+++ b/tests/unit/TestNetRemoteDataStreamingReactors.cxx
@@ -111,7 +111,7 @@ DataStreamReader::OnDone(const grpc::Status& status)
 }
 
 grpc::Status
-DataStreamReader::Await(uint32_t* numberOfDataBlocksReceived, DataStreamOperationStatus* operationStatus, std::vector<uint32_t>* lostDataBlockSequenceNumbers)
+DataStreamReader::Await(uint32_t* numberOfDataBlocksReceived, DataStreamOperationStatus* operationStatus, std::span<uint32_t> lostDataBlockSequenceNumbers)
 {
     std::unique_lock lock(m_readStatusGate);
 
@@ -129,7 +129,10 @@ DataStreamReader::Await(uint32_t* numberOfDataBlocksReceived, DataStreamOperatio
 
     *numberOfDataBlocksReceived = m_numberOfDataBlocksReceived;
     *operationStatus = m_data.status();
-    *lostDataBlockSequenceNumbers = m_lostDataBlockSequenceNumbers;
+
+    for (std::size_t i = 0; i < std::size(m_lostDataBlockSequenceNumbers); i++) {
+        lostDataBlockSequenceNumbers[i] = m_lostDataBlockSequenceNumbers[i];
+    }
 
     return m_status;
 }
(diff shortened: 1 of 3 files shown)


tests/unit/TestNetRemoteDataStreamingReactors.cxx
```diff
@@ -103,7 +113,7 @@ DataStreamReader::OnDone(const grpc::Status& status)
 }
 
 grpc::Status
-DataStreamReader::Await(uint32_t* numberOfDataBlocksReceived, DataStreamOperationStatus* operationStatus)
+DataStreamReader::Await(uint32_t* numberOfDataBlocksReceived, DataStreamOperationStatus* operationStatus, std::span<uint32_t> lostDataBlockSequenceNumbers)
```
abeltrano, line 116: I'm not sure I understand the usage of this variable. The `std::span` is being passed as a value copy, but it seems like this is intended to be used as an output variable instead. If that's true, this should be a reference instead (`std::span<uint32_t>&`). Have I misunderstood something here?

tests/unit/TestNetRemoteDataStreamingReactors.cxx
```diff
@@ -194,8 +205,10 @@ DataStreamReaderWriter::Await(uint32_t* numberOfDataBlocksReceived, Microsoft::N
         status.set_message("Timeout occurred while waiting for all operations to be completed");
         *m_readData.mutable_status() = std::move(status);
     }
+
     *numberOfDataBlocksReceived = m_numberOfDataBlocksReceived;
     *operationStatus = m_readData.status();
+    lostDataBlockSequenceNumbers = std::span<uint32_t>(std::data(m_lostDataBlockSequenceNumbers), std::size(m_lostDataBlockSequenceNumbers));
```
abeltrano, line 211: This is modifying the value passed to the function, then not used. Was `lostDataBlockSequenceNumbers` meant to be a reference such that this update's the caller's arg?

tests/unit/TestNetRemoteDataStreamingServiceClient.cxx
```diff
@@ -116,10 +116,12 @@ TEST_CASE("DataStreamDownload API", "[basic][rpc][client][remote][stream]")
 
         uint32_t numberOfDataBlocksReceived{};
         DataStreamOperationStatus operationStatus{};
-        const grpc::Status status = dataStreamReader.Await(&numberOfDataBlocksReceived, &operationStatus);
+        std::vector<uint32_t> lostDataBlockSequenceNumbers{};
+        const grpc::Status status = dataStreamReader.Await(&numberOfDataBlocksReceived, &operationStatus, lostDataBlockSequenceNumbers);
         REQUIRE(status.ok());
         REQUIRE(numberOfDataBlocksReceived == fixedNumberOfDataBlocksToStream);
         REQUIRE(operationStatus.code() == DataStreamOperationStatusCodeSucceeded);
+        REQUIRE(lostDataBlockSequenceNumbers.empty());
```
abeltrano, line 124: This will always be true since a copy of `lostDataBlockSequenceNumbers` is passed in.

tests/unit/TestNetRemoteDataStreamingReactors.cxx
corbin-phipps: Yes, should be a reference. Will update

PR #231, opened 2024-03-19
Add DataStreamPing unary RPC for checking server availability

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [X] Non-functional change

### Goals

Due to the nature of RPCs in gRPC, the only way to see if the remote server is unavailable is to try an RPC and check the status of the failure for status code 14 (UNAVAILABLE). Because the data streaming service uses streaming RPCs, this status code is not returned until operations are complete.

With this behavior, clients cannot check the availability of the server without trying the RPC and waiting for all operations to complete, which may be immediate if the server is unavailable or indefinite in successful streams. This is undesirable in cases where a disconnect occurs during a previous RPC call (server is unavailable) and the client wants to know exactly when the server is ready to successfully handle the next RPC.

Therefore, this PR adds a simple unary RPC called `DataStreamPing` that allows clients to quickly get a status from the server without request/response objects.

### Technical Details

* Added `DataStreamPing` unary RPC to `NetRemoteDataStreamingService`.
* Added a unit test for `DataStreamPing`.

### Test Results

All tests pass

### Reviewer Focus

None.

### Future Work

None.

### Checklist

### DIFF
--- a/src/common/service/NetRemoteDataStreamingService.cxx
+++ b/src/common/service/NetRemoteDataStreamingService.cxx
@@ -36,3 +36,14 @@ NetRemoteDataStreamingService::DataStreamBidirectional([[maybe_unused]] grpc::Ca
 
     return std::make_unique<Reactors::DataStreamReaderWriter>().release();
 }
+
+grpc::ServerUnaryReactor*
+NetRemoteDataStreamingService::DataStreamPing(grpc::CallbackServerContext* context, [[maybe_unused]] const google::protobuf::Empty* request, [[maybe_unused]] google::protobuf::Empty* response)
+{
+    const NetRemoteApiTrace traceMe{};
+
+    auto* reactor = context->DefaultReactor();
+    reactor->Finish(grpc::Status::OK);
+
+    return reactor;
+}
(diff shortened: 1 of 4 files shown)


src/common/service/NetRemoteDataStreamingService.cxx
```diff
@@ -36,3 +37,14 @@ NetRemoteDataStreamingService::DataStreamBidirectional([[maybe_unused]] grpc::Ca
 
     return std::make_unique<Reactors::DataStreamReaderWriter>().release();
 }
+
+grpc::ServerUnaryReactor*
```
abeltrano, line 41: I guess this has to return a reactor due to the `NetRemoteDataStreaming` impl type?

abeltrano (approved):

src/common/service/NetRemoteDataStreamingService.cxx
corbin-phipps: Yeah since I'm using the gRPC callback API, then the unary RPCs on the server side use `ServerUnaryReactor`. There is a bit more info here: https://github.com/grpc/proposal/blob/master/L67-cpp-callback-api.md

PR #236, opened 2024-03-20
Fix DataStreamDownload bug

### LINKED ISSUE #233: Bug: DataStreamDownload API test fails periodically
**Describe the bug**
The `DataStreamDownload API` test fails periodically with an assertion failure in the grpc code. I've only observed this on Linux and not Windows.

**Steps To Reproduce**

1. Run `netremote-test-unit "DataStreamDownload API"multiple times. It appears to happen more often with the [CI/CD GitHub action](https://github.com/microsoft/netremote/actions/workflows/cicd.yml), which can be manually triggered.

**Expected behavior**
The test passes.

**Actual behavior**
The test times out.

**Additional context**
This run captured one such failure:
https://github.com/microsoft/netremote/actions/runs/8347359873/job/22846799669

With relevant details regarding the assertion failure:
```bash
E0319 17:31:59.038000000  6588 grpcpp/support/callback_common.h:164] assertion failed: call_ == nullptr
```

corbin-phipps:
### Type

- [X] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR fixes a race condition bug where after the client canceled the `DataStreamDownload` RPC, a write from the server to the client failed before the server's `OnCancel()` callback could be invoked. The write failure led to the server completing the RPC. Then, once `OnCancel()` was invoked, the server tried to complete the RPC again, causing the crash.

Fixes #233 

### Technical Details

* Added `m_isCompleted` atomic boolean variable and used it to ensure that the server only completes the RPC once.
* Updated unit test for `DataStreamDownload` accordingly.

### Test Results

Ran all tests 20+ times in a row with no failures. Previously, I was able to repro the bug about every 10-15 test runs.

### Reviewer Focus

There may be a more elegant way to handle all the various race conditions in the server with a single sync mechanism rather than 2 atomic booleans, but this approach does seem to work.

### Future Work

None.

### Checklist

### DIFF
--- a/src/common/service/NetRemoteDataStreamingReactors.cxx
+++ b/src/common/service/NetRemoteDataStreamingReactors.cxx
@@ -145,7 +145,11 @@ DataStreamWriter::OnWriteDone(bool isOk)
 
     // Check for a failed status code from HandleWriteFailure since that invoked a final write, thus causing this callback to be invoked.
     if (m_writeStatus.code() == DataStreamOperationStatusCode::DataStreamOperationStatusCodeFailed) {
-        Finish(::grpc::Status::OK);
+        bool isCompletedExpected{ false };
+        if (m_isCompleted.compare_exchange_strong(isCompletedExpected, true, std::memory_order_relaxed, std::memory_order_relaxed)) {
+            Finish(grpc::Status::OK);
+        }
+
         return;
     }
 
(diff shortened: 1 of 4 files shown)


abeltrano:
> This PR fixes a race condition bug where after the client canceled the `DataStreamDownload` RPC, a write from the server to the client failed before the server's `OnCancel()` callback could be invoked. The write failure led to the server completing the RPC. Then, once `OnCancel()` was invoked, the server tried to complete the RPC again, causing the crash.

Very nice find 👍

> There may be a more elegant way to handle all the various race conditions in the server with a single sync mechanism rather than 2 atomic booleans, but this approach does seem to work.

I went through this carefully and I think your current logic is good. An alternative approach is to combine the `m_isCanceled` and `m_isCompleted` into a single `m_isFinished`, and then introduce a wrapper function for invoking `Finish`. However, with your current solution, having them separate makes clear why the operation was finished (and it's logged as such), so, I'd keep what you have.

abeltrano (approved):

PR #264, opened 2024-05-13
Add support to set network access server setting FT-based AKMs in hostapd

abeltrano:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

* Ensure fast-transition based AKMs work with hostapd.

### Technical Details

* Add string helper function to generate a random ascii string of desired length.
* Convert strings library from type interface to static and move previously inlined code into newly added source file `StringHelpers.cxx`.
* Add constants for valid range of FT R0 key holder identifiers (802.11r `dot11FTR0KeyHolderId`).
* Add helper to determine if `WpaKeyManagement` supports fast-transition.
* Add helper function to generate random network access server identifier.
* Add logic to automatically set the `nas_identifier` property to randomly generated value when setting AKMs that support FT.
* Remove `WpaCore.cxx` and move its contents to `WpaCore.hxx` since the function was `constexpr` anyway.
* Update `wpa_controller_test_unit` and utilities to
    - Use the CMake install runstate dir for WPA control path directory.
    - Truncate temp generated `hostapd.conf` file if exists to ensure correct control path is used when switching cmake configure presets between dev and release.

### Test Results

* All unit tests pass.
* `wpa_controller_test_unit` passes for both dev and release configurations.

### Reviewer Focus

* None

### Future Work

* Unit tests need to be added.

### Checklist


src/linux/wpa-controller/include/Wpa/ProtocolHostapd.hxx
```diff
@@ -300,16 +300,42 @@ inline constexpr std::array<WpaKeyManagement, 27> AllWpaKeyManagements = {
...
     std::to_underlying(WpaKeyManagement::FtIeee8021x) |
     std::to_underlying(WpaKeyManagement::FtIeee8021xSha384) |
     std::to_underlying(WpaKeyManagement::FtSae) |
     std::to_underlying(WpaKeyManagement::FtFilsSha256) |
     std::to_underlying(WpaKeyManagement::FtFilsSha384);
 
+/**
+ * @brief All valid WpaKeyManagement values supporting fast-transition (FT).
+ */
+static constexpr std::initializer_list<WpaKeyManagement> WpaKeyManagementFt = {
```
corbin-phipps, line 316: This is missing `FtPsk`

corbin-phipps (changes requested):

src/linux/wpa-controller/include/Wpa/ProtocolHostapd.hxx
abeltrano: The only one that we actually need right now ..

abeltrano: Added

PR #307, opened 2024-07-02
Add GET_CONFIG command

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR allows NetRemote to send the GET_CONFIG command to hostapd and parse the results.

### Technical Details

* Added `WpaCommandGetConfig` and `WpaResponseGetConfig`.
* Added `Hostapd::GetConfiguration()` method.
* Added `HostapdBssConfiguration` structure to store parsed result of GET_CONFIG.
* Added two helper conversion functions: `WpaKeyManagementFromPropertyValue()` and `WpaCipherFromPropertyValue()`.
* Added basic unit test.

### Test Results

Unit tests pass.

### Reviewer Focus

* The response of GET_CONFIG contains various property strings, and it is known that the string value for 'key_mgmt' may have multiple whitespace-separated values, so the string is parsed accordingly. Need to see if the other strings may also contain multiple values or just one.

### Future Work

Additional unit tests.

### Checklist


src/linux/wpa-controller/Hostapd.cxx
```diff
@@ -135,6 +137,19 @@ Hostapd::GetProperty(std::string_view propertyName)
     return propertyValue;
 }
 
+HostapdBssConfiguration
+Hostapd::GetConfiguration()
+{
+    static constexpr WpaCommandGetConfig GetConfigCommand;
```
abeltrano, line 143: nit: init braces
```suggestion
    static constexpr WpaCommandGetConfig GetConfigCommand{};
```
```suggestion
    static constexpr WpaCommandGetConfig GetConfigCommand;
```

src/linux/wpa-controller/ProtocolHostapd.cxx
```diff
@@ -85,3 +86,55 @@ Wpa::WpaPreSharedKeyPropertyKeyAndValue(const WpaPreSharedKey& wpaPreSharedKey)
...
+WpaKeyManagement
+Wpa::WpaKeyManagementFromPropertyValue(std::string_view wpaKeyManagementProperty) noexcept
+{
+    std::string wpaKeyManagementString(wpaKeyManagementProperty);
+    std::istringstream wpaKeyManagementStream(wpaKeyManagementString);
+    WpaKeyManagement result{ WpaKeyManagement::Unknown };
+
+    for (std::string wpaKeyManagement; wpaKeyManagementStream >> wpaKeyManagement;) {
+        if (wpaKeyManagement == "WPA-EAP") {
+            result = static_cast<WpaKeyManagement>(std::to_underlying(result) | std::to_underlying(WpaKeyManagement::Ieee8021x));
```
abeltrano, line 99: Converting from an enum's intergral/underlying type should normally be done with [`magic_enum::enum_cast`](https://github.com/Neargye/magic_enum/blob/master/doc/reference.md#enum_cast) instead, since it handles invalid inputs. 

So, you could replace `WpaKeyManagement result{ };` with `std::optional<WpaKeyManagement>`, assign the result of `magic_enum::enum_cast` to it, then check `result.has_value()` at the end.

Though I suspect I'm missing something here. Is there a reason why each one of these is assigning the bitwise-OR of the underlying value of `WpaKeyManagement::Unknown`?

src/linux/wpa-controller/include/Wpa/WpaCommandGetConfig.hxx
```diff
@@ -0,0 +1,58 @@
...
+     * @brief Construct a new WpaCommandGetConfig object.
+     */
+    constexpr WpaCommandGetConfig() :
+        WpaCommand(ProtocolWpa::CommandPayloadGetConfig)
+    {
+    }
+
+private:
+    std::unique_ptr<WpaResponseParser>
+    CreateResponseParser(const WpaCommand* command, std::string_view responsePayload) const override;
```
abeltrano, line 29: Could we add a comment block here?

abeltrano (approved):

src/linux/wpa-controller/ProtocolHostapd.cxx
corbin-phipps: `WpaKeyManagement::Unknown` is just `0U`, so that just an empty initialization. In this loop, I'm bitwise-OR'ing each WPA key management value that is found in the property value string, since the string could contain multiple values. I can see if I can still incorporate `magic_enum::enum_cast` though

abeltrano: Given my confusion, may also be helpful to add a comment block with the function prototype, and/or a comment at the start roughly describing the approach here.

corbin-phipps: Actually, I just realized my code doesn't work as I thought. I could have a `uint32_t` that is a bitwise-OR of various WPA key management values, but then casting that back to `WpaKeyManagement` wouldn't match any of the values. Perhaps I should store a vector of `WpaKeyManagement` instead

src/linux/wpa-controller/ProtocolHostapd.cxx
```diff
@@ -85,3 +86,55 @@ Wpa::WpaPreSharedKeyPropertyKeyAndValue(const WpaPreSharedKey& wpaPreSharedKey)
...
 }
+
+std::vector<WpaKeyManagement>
+Wpa::WpaKeyManagementFromPropertyValue(std::string_view wpaKeyManagementProperty) noexcept
+{
+    std::string wpaKeyManagementString(wpaKeyManagementProperty);
+    std::istringstream wpaKeyManagementStream(wpaKeyManagementString);
+    std::vector<WpaKeyManagement> wpaKeyManagements{};
+
+    for (std::string wpaKeyManagement; wpaKeyManagementStream >> wpaKeyManagement;) {
```
abeltrano, line 97: I should have mentioned this earlier, but I think there is value in separating the conversion of a single string to a single `WpaKeyManagement` value (ie. factor out the code within the for loop here into its own function, then just call that from here). This way we could write a test to validate the round-trip on the conversion is accurate, eg. 

```C++
for (auto wpaKeyManagement : AllWpaKeyManagements) {
    REQUIRE(WpaKeyManagementFromProperryValue(WpaKeyManagementPropertyValue(wpaKeyManagement)) == wpaKeyManagement);
}
```

This will ensure the corresponding conversion functions are at least consistent. Though, this is pretty trivial to change, so let's just keep what you have, and I can send a PR with that change tomorrow.

PR #310, opened 2024-07-11
Setup skeleton audit log

corbin-phipps:
### Type

- [ ] Bug fix
- [X] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

This PR sets up the foundation for the Wi-Fi state audit log.

### Technical Details

* Added a PLOG RollingFileAppender separate from the existing logging to be used for writing to the audit log file.
* Added audit log macros in the form of `AUDIT_LOG*`.
* Ensured that log files are actually present by specifying the path to `/usr/local/`.

### Test Results

Verified that the `AUDIT_LOGN` line added to `Main` shows up in `/usr/local/202407-LogNetRemote-audit.txt`.

### Reviewer Focus

I'm not sure if `/usr/local` is the best place, but omitting that and just using what is returned by `GetLogName` doesn't seem to create a log file anywhere.

### Future Work

Write Wi-Fi state and events to the audit log.

### Checklist

### DIFF
--- a/src/linux/server/Main.cxx
+++ b/src/linux/server/Main.cxx
@@ -35,21 +35,6 @@ using namespace Microsoft::Net::Remote;
 using namespace Microsoft::Net::Remote::Service;
 using namespace Microsoft::Net::Wifi;
 
-enum class LogInstanceId : int {
-    // Default logger is 0 and is omitted from this enumeration.
-    Console = 1,
-    File = 2,
-    Audit = 3,
-};
-
-#define AUDIT_LOGN LOG_(static_cast<int>(LogInstanceId::Audit), plog::none)
-#define AUDIT_LOGF LOG_(static_cast<int>(LogInstanceId::Audit), plog::fatal)
-#define AUDIT_LOGE LOG_(static_cast<int>(LogInstanceId::Audit), plog::error)
-#define AUDIT_LOGW LOG_(static_cast<int>(LogInstanceId::Audit), plog::warning)
-#define AUDIT_LOGI LOG_(static_cast<int>(LogInstanceId::Audit), plog::info)
-#define AUDIT_LOGD LOG_(static_cast<int>(LogInstanceId::Audit), plog::debug)
-#define AUDIT_LOGV LOG_(static_cast<int>(LogInstanceId::Audit), plog::verbose)
-
 namespace
 {
 /**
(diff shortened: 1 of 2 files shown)


src/linux/server/Main.cxx
```diff
@@ -89,7 +83,14 @@ main(int argc, char *argv[])
 
     // Create file and console log appenders.
     static plog::ColorConsoleAppender<plog::MessageOnlyFormatter> colorConsoleAppender{};
-    static plog::RollingFileAppender<plog::TxtFormatter> rollingFileAppender(logging::GetLogName("server").c_str());
+    std::string logFilePath = "/usr/local/";
```
abeltrano, line 86: FYI, no path prefix was previously added such that the current working directory would be used, which when run from systemd, is configurable and can be chosen to whatever we want instead of hard-coding it.

abeltrano (changes requested):
> I'm not sure if `/usr/local` is the best place, but omitting that and just using what is returned by `GetLogName` doesn't seem to create a log file anywhere.

The log file as currently used, will be created in the current working directory of the binary. This is configurable using the [`WorkingDirectory=`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#WorkingDirectory=) configuration directive in `[Service]` section of the `netremote-server` systemd unit file. I'd guess the binary doesn't have permission to write to that directory (`/usr/bin` or `/usr/local/bin`) since the service is started as a non-root user. So, the fix should be to set `WorkingDirectory` to something a regular user can access.

abeltrano (commented):
> > I'm not sure if `/usr/local` is the best place, but omitting that and just using what is returned by `GetLogName` doesn't seem to create a log file anywhere.

> The log file as currently used, will be created in the current working directory of the binary. This is configurable using the [`WorkingDirectory=`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#WorkingDirectory=) configuration directive in `[Service]` section of the `netremote-server` systemd unit file. I'd guess the binary doesn't have permission to write to that directory (`/usr/bin` or `/usr/local/bin`) since the service is started as a non-root user. So, the fix should be to set `WorkingDirectory` to something a regular user can access, like `/var/log`.

src/linux/server/Main.cxx
abeltrano, line 86: Let's change this to the standard logging directory, `/var/log`, for now:
```suggestion
    std::string logFilePath = "/var/log/";
```

abeltrano (approved):

PR #311, opened 2024-07-11
Log Wi-Fi state changes to the audit log

corbin-phipps:
### Type

- [ ] Bug fix
- [ ] Feature addition
- [X] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals

To log Wi-Fi state changes to the audit log. This is done at the neutral `Ieee80211*` layer in `AccessPointControllerLinux` for most of the `Set*` functions.

A debug ( `AUDITD` ) line is added to log which neutral type values will be attempted to be set, then after the values are set, the updated changes are validated. If successful, an info ( `AUDITI` ) line is added to log the successful changes. If unsuccessful (usually meaning the value retrieved by hostapd is not what was set), then a debug line is added to log the value that was retrieved.

### Technical Details

* Added several `AUDITD` and `AUDITI` lines in `AccessPointControllerLinux.cxx` as described above.
* Added various `Ieee80211*ToString()` functions to `Ieee80211.*xx`.

### Test Results

Verified that the following changes show up in the audit log:
* Setting operational state
* Setting PHY type
* Setting SSID

Have not been able to verify that setting AKM suites and pairwise cipher suites works.

Tested these changes with `netremote-cli wifi ap-enable wlo1 --phy 3 --akms 1027080 --sae password`, `netremote-cli wifi ap-disable wlo1`, and `netremote-cli wifi set-ssid wlo1 netremote2-ap-wlo1`.

### Reviewer Focus

None.

### Future Work

* Add WPA events to audit log (via `OnWpaEvent` callback).
* Add access point events (such as AP added, removed, etc.) to audit log (via `AccessPointManager`).
* Need to update `HostapdBssConfiguration` to use `std::vector<WpaCipher` for `WpaPairwiseCipher` and `RsnPairwiseCipher`. Currently, there is only a single `WpaCipher` for each.

### Checklist

### DIFF
--- a/src/common/net/wifi/core/Ieee80211.cxx
+++ b/src/common/net/wifi/core/Ieee80211.cxx
@@ -11,37 +11,7 @@
 
 namespace Microsoft::Net::Wifi
 {
-std::string
-Ieee80211SecurityProtocolToString(const Ieee80211SecurityProtocol& securityProtocol)
-{
-    switch (securityProtocol) {
-    case Ieee80211SecurityProtocol::Wpa:
-        return "Wpa";
-    case Ieee80211SecurityProtocol::Wpa2:
-        return "Wpa2";
-    case Ieee80211SecurityProtocol::Wpa3:
-        return "Wpa3";
-    default:
-        return "Unknown";
-    }
-}
-
-std::string
-Ieee80211FrequencyBandToString(const Ieee80211FrequencyBand& frequencyBand)
-{
-    switch (frequencyBand) {
-    case Ieee80211FrequencyBand::TwoPointFourGHz:
-        return "TwoPointFourGHz";
-    case Ieee80211FrequencyBand::FiveGHz:
-        return "FiveGHz";
-    case Ieee80211FrequencyBand::SixGHz:
-        return "SixGHz";
-    default:
-        return "Unknown"; 
-    }
-}
-
-std::string
+std::string_view
 Ieee80211AuthenticationAlgorithmToString(const Ieee80211AuthenticationAlgorithm& authenticationAlgorithm)
 {
     switch (authenticationAlgorithm) {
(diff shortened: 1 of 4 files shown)


src/common/net/wifi/core/Ieee80211.cxx
```diff
@@ -11,6 +11,149 @@
 
 namespace Microsoft::Net::Wifi
 {
```
abeltrano, line 13: It seems like all of the `*ToString` functions added here return the enumeration name, which is what `magic_enum::enum_name()` does. Is there any reason we can't just use that instead? If nothing else, could the functions be implemented with that instead?

The only exception is for the `Ieee80211AuthenticationAlgorithm` for which the `VendorSpecific` value exceeds `MAGIC_ENUM_RANGE_MAX`, thus preventing the use of `magic_enum::enum_name()` (and other magic_enum helpers).

src/common/net/wifi/core/Ieee80211.cxx
```diff
@@ -11,6 +11,149 @@
...
+std::string
```
abeltrano, line 14: This and others added in this PR should return `std::string_view` instead.

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -121,19 +124,35 @@ AccessPointControllerLinux::SetOperationalState(AccessPointOperationalState oper
...
         try {
             m_hostapd.Disable();
             status.Code = AccessPointOperationStatusCode::Succeeded;
+
+            // Validate that disabling the operational state succeeds.
+            AccessPointOperationalState actualOperationalState{};
+            auto getOperationalStateStatus = GetOperationalState(actualOperationalState);
+            if (getOperationalStateStatus.Code == AccessPointOperationStatusCode::Succeeded && actualOperationalState == AccessPointOperationalState::Disabled) {
+                AUDIT_LOGI << std::format("Operational state of AP {} set to 'disabled'", status.AccessPointId);
+            }
```
abeltrano: This code won't run if the call to `Disable` throws an exception. The `Hostapd` class signals errors by throwing a `HostapdException` instead of by return value or output parameter, so I think this will need to be accounted for here.

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -366,6 +448,18 @@ AccessPointControllerLinux::SetAkmSuites(std::vector<Ieee80211AkmSuite> akmSuite
...
 
+    // Validate that the AKM suites are successfully set by hostapd.
+    auto hostapdBssConfiguration = m_hostapd.GetConfiguration();
+    auto actualKeyManagements = hostapdBssConfiguration.WpaKeyMgmt;
+    if (std::equal(std::begin(wpaKeyManagements), std::end(wpaKeyManagements), std::begin(actualKeyManagements))) {
+        AUDIT_LOGI << std::format("AKM suites of AP {} set to {}", status.AccessPointId, akms);
+    } else {
+        AUDIT_LOGD << std::format("AKM suites of AP {} were not set. Current key_mgmt values:", status.AccessPointId);
+        for (const auto& actualKeyManagement : actualKeyManagements) {
+            AUDIT_LOGD << std::format("key_mgmt: {}", std::to_underlying(actualKeyManagement));
```
abeltrano: This will have one entry per actual key management value set. Should we instead create an aggregated string that has all of them so we can add a single audit log entry?

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -384,6 +478,20 @@ AccessPointControllerLinux::SetPairwiseCipherSuites(std::unordered_map<Ieee80211
...
+    std::string pairwiseCiphers{};
+    for (const auto& [securityProtocol, cipherSuites] : pairwiseCipherSuites) {
+        pairwiseCiphers += Ieee80211SecurityProtocolToString(securityProtocol);
+        pairwiseCiphers += ": ";
+        for (const auto& cipherSuite : cipherSuites) {
+            pairwiseCiphers += Ieee80211CipherSuiteToString(cipherSuite);
+            pairwiseCiphers += ",";
+        }
+        pairwiseCiphers.pop_back(); // Remove trailing comma
+        pairwiseCiphers += ";";
```
abeltrano, line 490: What about this instead?

```suggestion
        pairwiseCiphers.back() = ";';
```

abeltrano (changes requested):

src/common/net/wifi/core/Ieee80211.cxx
corbin-phipps: Oh good point, yeah that should work. I will remove these except for the auth algorithm one you mentioned.

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
corbin-phipps: That was intentional because I assumed that the state wouldn't change to 'disabled' if `Disable` fails/throws, so I put this in the `try` block, but maybe it's better to move this outside of the `try/catch`

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -366,6 +450,19 @@ AccessPointControllerLinux::SetAkmSuites(std::vector<Ieee80211AkmSuite> akmSuite
...
+    // Validate that the AKM suites are successfully set by hostapd.
+    auto hostapdBssConfiguration = m_hostapd.GetConfiguration();
+    auto actualKeyManagements = hostapdBssConfiguration.WpaKeyMgmt;
+    if (std::equal(std::begin(wpaKeyManagements), std::end(wpaKeyManagements), std::begin(actualKeyManagements))) {
+        AUDITI << std::format("AKM suites of AP {} set to {}", status.AccessPointId, akms);
+    } else {
+        std::string keyMgmtDebugLog = std::format("AKM suites of AP {} were not set. Current key_mgmt values: ", status.AccessPointId);
+        for (const auto& actualKeyManagement : actualKeyManagements) {
+            keyMgmtDebugLog += std::format("{} ", std::to_underlying(actualKeyManagement));
+        }
```
abeltrano, line 462: For future reference only, [`std::ostringstream`](https://en.cppreference.com/w/cpp/io/basic_ostringstream) should really be used for this and others building up a string instead as `std::string::operator+=` isn't super-efficient, whereas `std::ostringstream` was made for exactly this scenario. This way, you can also eliminate `std::format` here (and elsewhere) since the stream accepts integer types as well.

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -366,6 +450,19 @@ AccessPointControllerLinux::SetAkmSuites(std::vector<Ieee80211AkmSuite> akmSuite
         return status;
     }
 
+    // Validate that the AKM suites are successfully set by hostapd.
+    auto hostapdBssConfiguration = m_hostapd.GetConfiguration();
+    auto actualKeyManagements = hostapdBssConfiguration.WpaKeyMgmt;
+    if (std::equal(std::begin(wpaKeyManagements), std::end(wpaKeyManagements), std::begin(actualKeyManagements))) {
```
abeltrano, line 456: (future reference only, no need to fix now)

nit: Any time iterators are being used in a read-only fashion (as here), the `const` iterator versions should be used instead:

```suggestion
    if (std::equal(std::cbegin(wpaKeyManagements), std::cend(wpaKeyManagements), std::cbegin(actualKeyManagements))) {
```

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
```diff
@@ -394,6 +504,42 @@ AccessPointControllerLinux::SetPairwiseCipherSuites(std::unordered_map<Ieee80211
...
+    // Validate that the pairwise cipher suites were successfully set by hostapd.
+    auto hostapdBssConfiguration = m_hostapd.GetConfiguration();
+    auto wpaSecurityProtocol = hostapdBssConfiguration.Wpa;
+    bool isWpaSecurityProtocolSet = std::ranges::contains(pairwiseCipherSuitesHostapd, wpaSecurityProtocol, [](const auto& wpaCipher) {
+        return wpaCipher.first;
+    });
+    if (isWpaSecurityProtocolSet) {
+        AUDITI << std::format("WPA security protocol for AP {} set to {}", status.AccessPointId, std::to_underlying(wpaSecurityProtocol));
+    } else {
+        AUDITD << std::format("WPA security protocol for AP {} was not set. Current wpa value: {}", status.AccessPointId, std::to_underlying(wpaSecurityProtocol));
```
abeltrano, line 516: Using `std::to_underlying` here will put numerical enum value in the log. Wouldn't it be better to have the name instead, using the `*ToString()` function or `magic_enum::enum_name(wpaSecurityProtocol)`? Same for instances of `std::to_underlying` used below.

abeltrano (changes requested):

src/linux/net/wifi/core/AccessPointControllerLinux.cxx
corbin-phipps: Yeah, the name is probably better. My reasoning for using the numerical value was that it matches what gets returned/displayed by the hostapd CLI "GET_CONFIG" command (e.g. "wpa=2").

abeltrano (approved):

PR #348, opened 2025-10-14
Implement delayed enable and disable access point

wptestyyh:
### Type

- [ ] Bug fix
- [X] Feature addition
- [ ] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [ ] Non-functional change

### Goals
Implement gRPC api to enable or disable access point after specified seconds
### Technical Details
- Add ``WifiAccessPointTimedEnable`` and ``WifiAccessPointTimedDisable``
- Implement ``NetRemoteService::WifiAccessPointTimedEnable`` and ``NetRemoteService::WifiAccessPointTimedDisable``
- Implement unit test ``WifiAccessPointTimedEnable API`` and ``WifiAccessPointTimedDisable API``
### Test Results
- WifiAccessPointTimedDisable test passed
./netremote-test-unit "WifiAccessPointTimedDisable API"
===============================================================================
All tests passed (31 assertions in 1 test case)
- WifiAccessPointTimedEnable test passed
===============================================================================
All tests passed (31 assertions in 1 test case)

### Reviewer Focus
None
### Future Work
Publish 0.5.3 release

### Checklist

### DIFF
--- a/src/common/service/NetRemoteService.cxx
+++ b/src/common/service/NetRemoteService.cxx
@@ -711,11 +711,12 @@ NetRemoteService::WifiAccessPointTimedEnableImpl(std::string_view accessPointId,
     }
 
     // Create and store the timer thread
-    auto timerThread = std::make_shared<std::thread>([this, accessPointId = std::string(accessPointId), dot11AccessPointConfiguration, accessPointController, durationSeconds]() {
+    auto timerThread = std::make_shared<std::thread>([this, accessPointId = std::string(accessPointId), hasConfiguration = (dot11AccessPointConfiguration != nullptr), configurationCopy = dot11AccessPointConfiguration ? *dot11AccessPointConfiguration : Dot11AccessPointConfiguration{}, accessPointController, durationSeconds]() {
         std::this_thread::sleep_for(std::chrono::seconds(durationSeconds));
 
         // Enable the access point after the duration expires
-        auto result = WifiAccessPointEnableImpl(accessPointId, dot11AccessPointConfiguration, accessPointController);
+        const auto* configPtr = hasConfiguration ? &configurationCopy : nullptr;
+        auto result = WifiAccessPointEnableImpl(accessPointId, configPtr, accessPointController);
         if (result.code() != WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeSucceeded) {
             LOGW << std::format("Failed to automatically enable access point {} after {} seconds: {}",
                 accessPointId,
--- a/src/common/service/NetRemoteService.cxx
+++ b/src/common/service/NetRemoteService.cxx
@@ -1,12 +1,15 @@
 
 #include <algorithm>
+#include <chrono>
 #include <cstddef>
 #include <format>
+#include <future>
 #include <iterator>
 #include <memory>
 #include <ranges>
 #include <string>
 #include <string_view>
+#include <thread>
 #include <utility>
 #include <vector>
 
@@ -279,6 +282,18 @@ NetRemoteService::NetRemoteService(std::shared_ptr<NetworkManager> networkManage
 {
 }
 
+NetRemoteService::~NetRemoteService()
+{
+    // Wait for all timer threads to complete naturally
+    std::lock_guard<std::mutex> lock(m_threadsMutex);
+    if (m_timedEnableThread && m_timedEnableThread->joinable()) {
+        m_timedEnableThread->join();
+    }
+    if (m_timedDisableThread && m_timedDisableThread->joinable()) {
+        m_timedDisableThread->join();
+    }
+}
+
 std::shared_ptr<AccessPointManager>
 NetRemoteService::GetAccessPointManager() noexcept
 {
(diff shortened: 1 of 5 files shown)


src/common/service/NetRemoteService.cxx
```diff
@@ -642,6 +682,121 @@ NetRemoteService::WifiAccessPointDisableImpl(std::string_view accessPointId, std
...
+        std::lock_guard<std::mutex> lock(m_threadsMutex);
+        if (m_timedEnableThread && m_timedEnableThread->joinable()) {
+            wifiOperationStatus.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported);
+            wifiOperationStatus.set_message("A timed enable operation is already in progress");
+            return wifiOperationStatus;
+        }
+    }
+
+    // Create and store the timer thread
+    auto timerThread = std::make_shared<std::thread>([this, accessPointId = std::string(accessPointId), hasConfiguration = (dot11AccessPointConfiguration != nullptr), configurationCopy = dot11AccessPointConfiguration ? *dot11AccessPointConfiguration : Dot11AccessPointConfiguration{}, accessPointController, durationSeconds]() {
```
lewonhee: any indenting on this long line?

src/common/service/NetRemoteService.cxx
```diff
@@ -642,6 +682,121 @@ NetRemoteService::WifiAccessPointDisableImpl(std::string_view accessPointId, std
...
+            LOGI << std::format("Successfully automatically enabled access point {} after {} seconds",
+                accessPointId,
+                durationSeconds);
+        }
+    });
+
+    // Store the thread for management
+    {
+        std::lock_guard<std::mutex> lock(m_threadsMutex);
+        m_timedEnableThread = timerThread;
```
lewonhee, line 746: what if this thread is already created by other concurrent call? Can we just put whole logic creating thread and assigning it to member into this lock?

src/common/service/NetRemoteService.cxx
```diff
@@ -642,6 +682,121 @@ NetRemoteService::WifiAccessPointDisableImpl(std::string_view accessPointId, std
...
+    if (durationSeconds > MaxDurationSeconds) {
+        wifiOperationStatus.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeInvalidParameter);
+        wifiOperationStatus.set_message(std::format("Duration {} seconds exceeds maximum allowed duration of {} seconds", durationSeconds, MaxDurationSeconds));
+        return wifiOperationStatus;
+    }
+
+    // Check if a timed enable operation is already running
+    {
+        std::lock_guard<std::mutex> lock(m_threadsMutex);
+        if (m_timedEnableThread && m_timedEnableThread->joinable()) {
```
lewonhee, line 710: How about DisableThread? If DisableThread exists, should we block this? Also, can this service manage multiple APs? If yes, this breaks that design since `m_timedEnableThread` is globally managed.

src/common/service/NetRemoteService.cxx
```diff
@@ -279,6 +282,18 @@ NetRemoteService::NetRemoteService(std::shared_ptr<NetworkManager> networkManage
 {
 }
 
+NetRemoteService::~NetRemoteService()
+{
+    // Wait for all timer threads to complete naturally
```
lewonhee: I don't like waiting thread at destructor. Copilot suggests the following ways as best practices. Could you adopt one of this?

1. Use Cancellation/Stop Signals
Instead of just waiting for the thread to finish, you can use a cancellation flag (like std::atomic<bool> m_stopRequested) that threads check periodically. When shutting down, set this flag and then notify the threads (sometimes with a condition variable). Threads should exit cleanly and promptly when the flag is set.

Example:
```
class NetRemoteService {
    std::atomic<bool> m_stopRequested{false};
    // ...
};

void timerThreadFunc(/*...*/) {
    for (uint32_t i = 0; i < durationSeconds; ++i) {
        if (m_stopRequested) return; // Exit immediately if shutdown requested
        std::this_thread::sleep_for(std::chrono::seconds(1));
    }
    // Perform enable/disable...
}
```
Then in your destructor:

```
m_stopRequested = true;
if (m_timedEnableThread && m_timedEnableThread->joinable()) {
    m_timedEnableThread->join();
}
```
2. Use Condition Variables for Early Wake-up
If your thread is sleeping for long durations, use a std::condition_variable so you can wake it early during shutdown.

Example:

```
std::condition_variable m_cv;
std::mutex m_cvMutex;

void timerThreadFunc(uint32_t durationSeconds) {
    std::unique_lock<std::mutex> lk(m_cvMutex);
    if (m_cv.wait_for(lk, std::chrono::seconds(durationSeconds), [this]{ return m_stopRequested; })) {
        // Stop requested, exit early
        return;
    }
    // Perform enable/disable...
}
```
During shutdown, set the stop flag and call m_cv.notify_all() to wake the thread.

3. Avoid Blocking the Destructor
If possible, detach the threads or use a thread pool with proper lifecycle management, so the destructor doesn't block for long. Only join for a short time or log if threads take too long.

4. Design for Idempotent Operations
Make sure your timed enable/disable operations are safe to cancel or abort midway if shutdown is requested. This prevents partial/invalid states.

lewonhee (changes requested):

src/common/service/NetRemoteService.cxx
wptestyyh: Understand the concern.
The service almost never stop, and the thread only have a simple action that enables/disables the access point after the specified seconds. I limit the longest seconds to 60. That is why I don't implement extra logic that can quit the thread earlier. 
If you insist that is not enough and needs the cancel mechanism, I can use conditional_variable to do so.

src/common/service/NetRemoteService.cxx
wptestyyh: nope, it can only manage a simple AP right now. Disable thread and enable thread are separated, this implementation is only for a simple purpose that test client can disable the access point in the middle of the standby. Right now, I don't see we need mix up delayed enable and disable in the test. Even if it is needed, the service won't limit how they are mixed up to provide the full flexibility that the test client can choose

src/common/service/NetRemoteService.cxx
wptestyyh: I did format the change (ctrl+k and ctrol+f), but vs does nothing. I will break the long time manually in next iteration

src/common/service/NetRemoteService.cxx
wptestyyh: Understand the concern.
The service almost never stop, and the thread only have a simple action that enables/disables the access point after the specified seconds. I limit the longest seconds to 600 (10 minutes). That is why I don't implement extra logic that can quit the thread earlier.
If you insist that is not enough and needs the cancel mechanism, I can use conditional_variable to do so.

src/common/service/NetRemoteService.cxx
wptestyyh: This is automatically format by .clang-format. Even if I manually break the long ling, it will be changed back to a single long time automatically when I save the file.  ".clang-format" settings may need be updated for this, but I won't do it in this change.

src/common/service/NetRemoteService.cxx
wptestyyh: We only allow a single enable thread to run at a time. But I do see there may be a race if other concurrent call happens that may start multiple enable thread. I will fix it in next iteration.

src/common/service/NetRemoteService.cxx
wptestyyh: Nope, netremote server is expected to only manage single AP right now. Disable and enable thread are completely separated

src/common/service/NetRemoteService.cxx
wptestyyh: Fixed with solution 1

lewonhee (approved):

wptestyyh:
The linux build is broken because of out of diskspace that is not the issue of the code change.
The windows build is https://mirror.msys2.org/msys/x86_64/msys2-runtime-3.4.10-4-x86_64.pkg.tar.zst has been deleted. I am trying to upgrade vcpkg submodule to fix it. No fix ETA yet

PR #350, opened 2025-10-22
Fix the wrong check of running thread

wptestyyh:
### Type

- [X] Bug fix
- [ ] Feature addition
- [ ] Feature update
- [ ] Documentation
- [ ] Build Infrastructure

### Side Effects

- [ ] Breaking change
- [x] Non-functional change

### Goals
std::thread::joinable() is not the same as running. The thread can be joinable but already finished. The second call will be failed always if we use joinable to check if the thread is running.

### Technical Details
- Create flag ``m_timedEnableRunning`` and ``m_timedDisableRunning`` to check if the thread has finished
- Add new test ``Succeeds when a timed enable operation is finished`` and ``Succeeds when a timed disable operation is finished``

### Test Results
- /netremote-test-unit "WifiAccessPointTimedEnable API"
===============================================================================
All tests passed (38 assertions in 1 test case)
./netremote-test-unit "WifiAccessPointTimedDisable API"
===============================================================================
All tests passed (40 assertions in 1 test case)

### Reviewer Focus
None

### Future Work
Publish the fix in 0.5.3

### Checklist

### DIFF
--- a/src/common/service/NetRemoteService.cxx
+++ b/src/common/service/NetRemoteService.cxx
@@ -707,11 +707,19 @@ NetRemoteService::WifiAccessPointTimedEnableImpl(std::string_view accessPointId,
     // Right now, this service only supports managing singale access point.
     {
         std::lock_guard<std::mutex> lock(m_threadsMutex);
-        if (m_timedEnableThread && m_timedEnableThread->joinable()) {
+        if (m_timedEnableRunning.load()) {
             wifiOperationStatus.set_code(WifiAccessPointOperationStatusCode::WifiAccessPointOperationStatusCodeOperationNotSupported);
             wifiOperationStatus.set_message("A timed enable operation is already in progress");
             return wifiOperationStatus;
         }
+
+        // Join previous thread if it exists and is joinable
+        if (m_timedEnableThread && m_timedEnableThread->joinable()) {
+            m_timedEnableThread->join();
+        }
+
+        // Set the running flag before creating the thread
+        m_timedEnableRunning = true;
 
         // Create and store the timer thread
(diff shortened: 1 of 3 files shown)


src/common/service/NetRemoteService.cxx
```diff
@@ -707,12 +707,20 @@ NetRemoteService::WifiAccessPointTimedEnableImpl(std::string_view accessPointId,
...
             return wifiOperationStatus;
         }
 
+        // Join previous thread if it exists and is joinable
+        if (m_timedEnableThread && m_timedEnableThread->joinable()) {
+            m_timedEnableThread->join();
+        }
+
+        // Set the running flag before creating the thread
+        m_timedEnableRunning = true;
```
lewonhee, line 722: what if thread object creation fail? This will be left as false.

src/common/service/NetRemoteService.cxx
```diff
@@ -773,12 +785,20 @@ NetRemoteService::WifiAccessPointTimedDisableImpl(std::string_view accessPointId
...
             wifiOperationStatus.set_message("A timed disable operation is already in progress");
             return wifiOperationStatus;
         }
 
+        // Join previous thread if it exists and is joinable
+        if (m_timedDisableThread && m_timedDisableThread->joinable()) {
+            m_timedDisableThread->join();
+        }
+
+        // Set the running flag before creating the thread
```
lewonhee, line 799: same as above

lewonhee (changes requested):

src/common/service/NetRemoteService.cxx
wptestyyh: That is rare error case that netremote doesn't handle. If it happens, netremote service will crash and stop

